PR #24, opened 2019-08-08
Add GPU support 

jacobtomlinson:
This PR aims to add GPU support for ECS cluster which have GPU instances. 

Tasks:
- Separate fargate toggle into fargate scheduler and fargate worker (you could have a fargate scheduler with EC2 workers)
- Add new kwarg `worker_gpu` to specify how many GPUs each worker gets
- Use correct docker images depending on GPU/noGPU work
- Launch `dask-cuda-worker` where appropriate

This has also exposed some issues with EC2 based clusters vs Fargate clusters surrounding networking which I am aiming to tackle here too.


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -948,8 +998,17 @@ def _task_role_name(self):
...
                     "memoryReservation": self._worker_mem,
+                    "resourceRequirements": resource_requirements,
                     "essential": True,
-                    "command": ["dask-worker"],
+                    "command": [
+                        "dask-cuda-worker" if self._worker_gpu else "dask-worker",
+                        "--nthreads",
+                        "{}".format(int(self._worker_cpu / 1024)),
+                        "--memory-limit",
+                        "{}GB".format(int(self._worker_mem / 1024)),
```
mrocklin: You can also use `MB` or `MiB` here if that's more convenient

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -541,14 +565,29 @@ def __init__(
...
+            self.config.get("fargate_workers", False)
+            if self._fargate_workers is None
+            else self._fargate_workers
         )
         self._tags = self.config.get("tags", {}) if self._tags is None else self._tags
+        self._worker_gpu = (
+            self.config.get("worker_gpu")
+            if self._worker_gpu is None
+            else self._worker_gpu
+        )  # TODO Detect whether cluster is GPU capable
```
mrocklin: How do people select which kind of GPU they want?  Is this handled in some other configuration definition somewhere?

jacobtomlinson: They do when they create the ECS cluster, which is out of scope for this library.

dask_cloudprovider/providers/aws/ecs.py
jacobtomlinson: These values are used in multiple places and those places are less flexible 😄

mrocklin: I'm not suggesting that we use different units for `self._worker_mem`

mrocklin:
```suggestion
                        "{}MB".format(int(self._worker_mem)),
```

jacobtomlinson: Ohh I see 😄

jacobtomlinson:
Tested this from an EC2 instance within the same VPC as a GPU ECS cluster and everything seems to work well. Merging this now and will continue to iterate on it, for example #25 adds support for out of VPC cluster managers using fargate schedulers.

mmccarty:
Thanks @jacobtomlinson Looking forward to trying this out!

PR #25, opened 2019-08-17
Extend GPU support to also support a Fargate scheduler

jacobtomlinson:
This PR extends #24 to include changes from dask/distributed#2963 which allow you to specify a fargate scheduler with an external IP and EC2 workers with GPUs that use internal IPs.


jacobtomlinson:
My PR dask/distributed#2963 is now merged but we need to hold this until the next distributed release so we can bump the minimum version.

mrocklin:
Thoughts on working from master for a while?  We had a distributed release just yesterday.  We could start to do more frequent point releases for a while, maybe next Friday?

jacobtomlinson:
I'm not in a huge rush for this PR. It's more for completeness. 

More frequent point releases could be nice though. But looking back through it seems to average every few weeks anyway.

PR #26, opened 2019-08-20
Move all default values into config file

### LINKED ISSUE #3: Uncomment the config file and use those defaults
We should be able to safely uncomment the config file in cloud.yaml .  Dask will both use these values as baseline defaults, which should stop us from having to double-specify them in sourcecode like here:

```python
        self._scheduler_cpu = (
            self.config.get("scheduler_cpu", 1024)  # <-- default now in two places
            if self._scheduler_cpu is None
            else self._scheduler_cpu
        )
```

and it will automatically comment out the user's version as it includes it in `~/.config/dask/cloud.yaml` by default.

jacobtomlinson:
Simplify code by moving all default values into the config file.

Closes #3

### DIFF
--- a/dask_cloudprovider/cloudprovider.yaml
+++ b/dask_cloudprovider/cloudprovider.yaml
@@ -1,21 +1,32 @@
-cloudprovider: {}
+cloudprovider:
 
-  # ecs: {}
-  #   # scheduler_cpu: 1024  # Millicpu (1024ths of a CPU core)
-  #   # scheduler_mem: 4096  # Memory in MB
-  #   # worker_cpu: 4096  # Millicpu (1024ths of a CPU core)
-  #   # worker_mem: 16384  # Memory in MB
-  #   # n_workers: 2
+  ecs:
+    fargate_scheduler: False  # Use fargate mode for the scheduler
+    fargate_workers: False  # Use fargate mode for the workers
+    scheduler_cpu: 1024  # Millicpu (1024ths of a CPU core)
+    scheduler_mem: 4096  # Memory in MB
+    worker_cpu: 4096  # Millicpu (1024ths of a CPU core)
+    worker_mem: 16384  # Memory in MB
+    worker_gpu: 0  # Number of GPUs for each worker
+    n_workers: 0  # Number of workers to start the cluster with
+    scheduler_timeout: '5 minutes'  # Length of inactivity to wait before closing the cluster
 
-  #   # image: 'daskdev/dask:1.2.0'
-  #   # cluster_name_template: 'dask-{uuid}'  # Template to use when creating a cluster
-  #   # cluster_arn: null  # ARN of existing ECS cluster to use (if not set one will be created)
-  #   # execution_role_arn: null  # Arn of existing execution role to use (if not set one will be created)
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
@mrocklin could you give this a quick review to check it satisfies #3?

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -1,21 +1,32 @@
...
-  #   # scheduler_cpu: 1024  # Millicpu (1024ths of a CPU core)
-  #   # scheduler_mem: 4096  # Memory in MB
-  #   # worker_cpu: 4096  # Millicpu (1024ths of a CPU core)
-  #   # worker_mem: 16384  # Memory in MB
-  #   # n_workers: 2
+  ecs:
+    fargate_scheduler: False  # Use fargate mode for the scheduler
+    fargate_workers: False  # Use fargate mode for the workers
+    scheduler_cpu: 1024  # Millicpu (1024ths of a CPU core)
+    scheduler_mem: 4096  # Memory in MB
```
mrocklin: From a Dask perspective it's more natural to me to use text here, like `4096 MB`.  However, I suspect that you've chosen this approach because it's more familiar to AWS users.

mrocklin:
Everything looks good to me

dask_cloudprovider/cloudprovider.yaml
jacobtomlinson: I chose this because the AWS API requires an integer value in megabytes (apart from the few times it wants a string of the integer ... _grumble, grumble_). If you think it would be more natural to have a string here following the dask approach I could look at parsing it. Are there any utils in distributed for doing this?

mrocklin:
```python
In [1]: from dask.utils import parse_bytes

In [2]: parse_bytes("4096 MB")
Out[2]: 4096000000

In [3]: parse_bytes?
Signature: parse_bytes(s)
Docstring:
Parse byte string to numbers

>>> parse_bytes('100')
100
>>> parse_bytes('100 MB')
100000000
>>> parse_bytes('100M')
100000000
>>> parse_bytes('5kB')
5000
>>> parse_bytes('5.4 kB')
5400
>>> parse_bytes('1kiB')
1024
>>> parse_bytes('1e6')
1000000
>>> parse_bytes('1e6 kB')
1000000000
>>> parse_bytes('MB')
1000000
>>> parse_bytes('5 foos')  # doctest: +SKIP
ValueError: Could not interpret 'foos' as a byte unit
File:      ~/workspace/dask/dask/utils.py
Type:      function
```

mrocklin: See also it's friends, `format_bytes`, `parse_timedelta`, `format_timedelta`

jacobtomlinson: Ah nice ok. I'd used `format_bytes` before but didn't know about `parse_bytes`.

jacobtomlinson: I'm going to get this merged and have raised #27 to keep track of your comments here.

PR #41, opened 2019-10-16
Adds boto client options

### LINKED ISSUE #30: Add kwargs to boto3 client options
The `ECSCluster` and `FargateCluster` classes use `boto3` to communicate with AWS. [Currently clients for each service are constructed and use default settings.](https://github.com/dask/dask-cloudprovider/blob/00ffbc649cc554802c86bd44881800153000e5dc/dask_cloudprovider/providers/aws/ecs.py#L719-L727) This results in `boto3` detecting information like access keys and regions from the environment or dotfiles.

It would be good to extend the cluster managers to optionally take some of the configuration options for `boto3` clients and pass them through, this would enable users to configure their keys, regions, etc at run time.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html#boto3.session.Session.client

I would imagine the following kwargs would fit most common use cases:

- `aws_access_key_id`
- `aws_secret_access_key`
- `region_name`

jrbourbeau:
This PR adds `aws_access_key_id`, `aws_secret_access_key`, and `region_name` keyword arguments to `ECSCluster.__init__` which are then used to override defaults when creating `botocore` service clients.

Closes #30

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -523,6 +523,9 @@ class ECSCluster(SpecCluster):
         environment=None,
         tags=None,
         skip_cleanup=None,
+        aws_access_key_id=None,
+        aws_secret_access_key=None,
+        region_name=None,
         **kwargs
     ):
         self._clients = None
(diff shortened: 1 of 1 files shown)


jrbourbeau:
I added these three options explicitly as I suspect they're commonly used. However, if we think several of the other client options should also be supported, we could instead have a `boto_client_options` (or some other name) parameter. I.e. something like:

```python
ECSCluster(...,
           boto_client_options={"aws_access_key_id": ..., "aws_secret_access_key": ...})
```

@jacobtomlinson do you have a preference?

jacobtomlinson:
I think I prefer the way you've implemented it here.

Thanks for doing this!

PR #42, opened 2019-10-17
Add address timeout config

kylejn27:
Adds a configuration for finding the task (mostly for the scheduler) address timeout. A 30 second timeout is too short for when you add in extra conda or pip packages. The user should be able to extend this time to as long as they require


jacobtomlinson:
This is great thanks!

Issue #34, opened 2019-09-16
Add sleep on GetLogEvents to prevent Rate Limit Exceeded


kylejn27:
GetLogEvents has a "10 requests per second per account per Region" limit. https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html

https://github.com/dask/dask-cloudprovider/blob/master/dask_cloudprovider/providers/aws/ecs.py#L270 should pause before polling for the next batch of logs

jacobtomlinson:
Closed in #44

PR #44, opened 2019-10-21
adds exponential backoff to get_log_events

kylejn27:
**Fixes**
https://github.com/dask/dask-cloudprovider/issues/34

**What**
Adds exponential backoff to get_log_events along with a warning message that this is happening. Max backoff is set at 5 seconds

Adapted from:
https://aws.amazon.com/blogs/messaging-and-targeting/how-to-handle-a-throttling-maximum-sending-rate-exceeded-error/


jacobtomlinson:
Thanks for implementing this.

Issue #53, opened 2020-01-17
Launching a FargateCluster with less than 1024 CPUs has wrong number of nthreads


gvelchuru:
I was launching a FargateCluster with 512 CPUs (.5 vCPUs) and I noticed the nthreads was 0 (which defaults to 2 in Fargate) -- I think this line in the FargateCluster constructor is the offender:

 "--nthreads", "{}".format(int(self._cpu / 1024)),

gvelchuru:
I have opened #54 as a potential solution.

jacobtomlinson:
Thanks for raising this. 

While your fix in #54 seems reasonable I would generally advise against giving a worker less then 1 CPU core.

Perhaps @mrocklin could advise whether there may be some trouble with doing this?

mrocklin:
No trouble technically.  Dask is usually used in compute heavy tasks
though, so this is rare and feels a bit odd.  I'm sure that valid use cases
exist though.

jacobtomlinson:
Thanks @mrocklin and thanks @gvelchuru for putting in a fix.

I'd be curious to hear more about your use cases where you are using <1 vCPU. Just for context.

gvelchuru:
I had a task that was memory-bound rather than CPU-bound, so I found it better to launch many instances with 1GB ram and .256GB CPU. Not sure if this is a one-off case though, since I ended up optimizing the graph and making the task CPU-bound.

PR #54, opened 2020-01-17
Fix nthreads in situations where cpu is less than 1

gvelchuru:
For issue #53

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -366,7 +366,7 @@ class Worker(Task):
                 "--name",
                 str(self.name),
                 "--nthreads",
-                "{}".format(int(self._cpu / 1024)),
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -366,7 +366,7 @@ def __init__(self, scheduler: str, cpu: int, mem: int, gpu: int, **kwargs):
                 "--name",
                 str(self.name),
                 "--nthreads",
-                "{}".format(int(self._cpu / 1024)),
+                "{}".format(int(self._cpu / 1024) if self._cpu >= 1024 else 1),
```
jrbourbeau: Thanks @gvelchuru! One small comment: could you switch this to something like:

```python
min(int(self._cpu / 1024), 1)
```

which is more concise

gvelchuru:
Fixed it!

PR #57, opened 2020-01-27
Improve GPU documentation

jacobtomlinson:
Add more detail to the GPU section to describe what happens as a result of setting the `worker_gpu` kwarg.


jakirkham:
Thanks Jacob! 😄

Issue #49, opened 2019-12-19
Required permissions policies to attach to IAM user


filippo82:
Hi all,

I am fairly new to the AWS world and I've decided to reproduce the results in @jacobtomlinson [blog post](https://medium.com/rapids-ai/getting-started-with-rapids-on-aws-ecs-using-dask-cloud-provider-b1adfdbc9c6e#cid=av01_so-twit_en-us). To do so, I first had to create a new IAM user (I thought it would be better to create a specific user to test dask-cloudprovider) without attaching any specific permissions policy. Then, after trying to execute `cluster = FargateCluster(n_workers=1)`, I experienced the first permissions issue:

> AccessDeniedException: An error occurred (AccessDeniedException) when calling the ListClusters operation: User: arn:aws:iam::707249732839:user/filippoawstest is not authorized to perform: ecs:ListClusters on resource: *

I *think* I solved this adding the AWS managed policy **AmazonECS_FullAccess**.

The second issue I had was:

> ClientError: An error occurred (AccessDenied) when calling the ListRoleTags operation: User: arn:aws:iam::707249732839:user/filippoawstest is not authorized to perform: iam:ListRoleTags on resource: role AWSServiceRoleForECS

I *think* I solved this adding the AWS managed policy **IAMFullAccess**.

The third issue I had was:

> An error occurred (UnauthorizedOperation) when calling the CreateTags operation: You are not authorized to perform this operation. Encoded authorization failure message: LONG_MESSAGE

After decoding the error message (`aws decode-authorization-message --encoded-message LONG_MESSAGE`), I believe it was related to this action `'action': 'ec2:CreateTags`. Again, I *think* I solved this adding the AWS managed policy **CloudWatchLogsReadOnlyAccess**.

After adding these policies, `cluster = FargateCluster(n_workers=1)` succeeded and produced this message:

> /work/bfilippo/miniconda3/envs/aws/lib/python3.7/contextlib.py:119: UserWarning: Creating your cluster is taking a surprisingly long time. This is likely due to pending resources on AWS. Hang tight! next(self.gen)

Then, the remaining of the simple dask array example worked (I could see that through the dashboard too):
```
import dask.array as da
arr = da.random.random((1000, 1000, 10000), chunks=(100, 100, 1000))
arr = arr.mean().persist()
arr.compute()
```

I didn't play with the ECS Cluster part.

**So my question is: Is there a better set of policies to attach to the user?**

Cheers.
-Filippo

[...]

PR #69, opened 2020-02-28
provide minimal IAM policy for creating FargateCluster

rsignell-usgs:


jacobtomlinson:
Thanks for this @rsignell-usgs!

I've moved this further down in the docs and also added a second example for folks who want to specify their own iam roles, subnets, etc which should require fewer permissions.

jacobtomlinson:
Closed #49

jkmacc-LANL:
This is awesome!  Thanks, @rsignell-usgs!  @jacobtomlinson, what is meant by “specify ... themselves” are In the minimal example?  I’m anticipating that many folks looking here for a quick zero-to-dask solution  (myself included) may not know what this means (myself included).

jacobtomlinson:
@jkmacc-LANL are you referring to these lines?

> To create a ``FargateCluster`` the cluster manager will need to various AWS resources ranging from IAM roles to VPCs to ECS tasks. Depending on your use case you may want the cluster to create all of these for you, or you may wish to specify them youself ahead of time.

jkmacc-LANL:
@jacobtomlinson I'm just thinking that, without visiting with you and @rsignell-usgs at the Dask workshop last month, I wouldn't know what to do with that block of JSON, even though you've specified it for me.  Is there a good resource or link you use to understand IAM policies and how to apply them, particularly for users of this library?

jacobtomlinson:
@jkmacc-LANL documentation [like this](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_create.html) should be useful to you. Perhaps we should also link that page form the documentation, although it was the top hit for "creating IAM policies" on Google.

I have made some assumptions that it would be good to check though:

- Individuals signing up for their own AWS account will likely be using the default administrator role with all of the required permissions.
- Institutions providing accounts to staff will likely have a process for requesting permissions. They will probably appreciate being provided a JSON policy like this.
- Folks setting up their own multi-user systems like Pangeo should have some understanding of AWS and should recognise an IAM policy and what to do with it.

These assumptions may be wrong, or I may have missed some use cases. Any input and feedback would be much appreciated!

PR #75, opened 2020-03-16
Todrabas devel

drabastomek:
Updating the requirements.txt to pass CI


drabastomek:
Will check into the main PR

PR #85, opened 2020-04-19
Support avoiding use of a public IP with Fargate

joeschmid:
Previously, a Fargate cluster would always use a public IP. For organizations with security and compliance constraints, they would need to use Fargate with a private subnet in a VPC that isn't reachable from the public Internet.

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -73,6 +73,12 @@ class Task:
         Configurable timeout in seconds for finding the task IP from the
         cloudwatch logs.
 
+    name: str (optional)
+        Name for used for the --name argument to dask-worker command line
+
+    use_private_ip: bool (optional)
+        Whether to use a private IP (if True) or public IP (if False). Default is False.
+
     kwargs:
         Any additional kwargs which may need to be stored for later use.
 
@@ -137,7 +143,7 @@ class Task:
 
     @property
     def _use_public_ip(self):
-        return not self._use_private_ip
+        return self.fargate and not self._use_private_ip
 
     async def _is_long_arn_format_enabled(self):
         async with self._client("ecs") as ecs:
@@ -521,7 +527,7 @@ class ECSCluster(SpecCluster):
 
         Default ``False``.
     use_private_ip: bool (optional)
-        Avoid using a public IP address, e.g. with a private subnet VPC.
+        If True, use a private IP, if False use public IPs.
 
         Default ``False``.
     **kwargs: dict
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -745,7 +745,7 @@ class ECSCluster(SpecCluster):
             "environment": self._environment,
             "tags": self.tags,
             "find_address_timeout": self._find_address_timeout,
-            "use_private_ip": self._use_private_ip
+            "use_private_ip": self._use_private_ip,
         }
         scheduler_options = {
             "task_definition_arn": self.scheduler_task_definition_arn,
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -135,14 +137,7 @@ async def _():
...
     def _use_public_ip(self):
-        # Fargate needs public IP for image pull, EC2 doesn't support public IP, therefore
-        # we will assume for now that we will use a public IP when in Fargate mode and not
-        # when in EC2 mode.
-
-        # TODO Fargate can also use a NAT to pull the image so we could allow this to be false
-        # when in Fargate provided there is a NAT
-
-        return self.fargate
+        return not self._use_private_ip
```
martindurant, line 140: should be `self.fargate and not self._use_private_ip`, no? Or otherwise, should be clear what are the right defaults for ec2 versus fargate (perhaps in init rather than here).

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -524,6 +519,10 @@ class ECSCluster(SpecCluster):
         Skip cleaning up of stale resources. Useful if you have lots of resources
         and this operation takes a while.
 
+        Default ``False``.
+    use_private_ip: bool (optional)
+        Avoid using a public IP address, e.g. with a private subnet VPC.
```
martindurant, line 524: This text could be clearer: 
if `True`, use private PRs, if `False` use public IPs. 
Must be `True` for ec2, but for fargate should be set to `True` when running in a private VPC.

(I don't know if the words above are actually correct)

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -120,6 +121,7 @@ def __init__(
         self.environment = environment or {}
         self.tags = tags
         self._find_address_timeout = find_address_timeout
+        self._use_private_ip = use_private_ip
```
martindurant, line 124: Other parameters seem to have a one-line description in the docstring here as well

joeschmid:
Thanks @martindurant for the feedback. Much appreciated!

dask_cloudprovider/providers/aws/ecs.py
joeschmid: Thanks, I agree with that. Updated to the logic you provided.

dask_cloudprovider/providers/aws/ecs.py
joeschmid: I like your clarification. I'm not certain on the ec2 behavior (only Fargate) so I avoided putting that in the description, but used your phrasing for the rest -- definitely more clear.

dask_cloudprovider/providers/aws/ecs.py
joeschmid: Thanks, fixed! (And noticed `name` didn't have a description so added that, too.)

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -73,6 +73,12 @@ class Task:
         Configurable timeout in seconds for finding the task IP from the
         cloudwatch logs.
 
+    name: str (optional)
+        Name for the task. Currently used for the --namecommand line argument to dask-worker.
+
+    use_private_ip: bool (optional)
+        Whether to use a private IP (if True) or public IP (if False). Default is False.
```
martindurant, line 80: This is still for fargate only, though, no? On ec2, this arg makes no difference.

joeschmid: That's correct. After your comment and thinking about it more, I changed this to `fargate_use_public_ip` to clarify that it will only have an effect for Fargate and not ec2. (Sorry for the back-and-forth, but thanks for bearing with me!)

martindurant:
OK! :)

PR #91, opened 2020-04-24
Allow extra command line args to scheduler & worker

joeschmid:
In order to support TLS encryption, we need to be able to pass command line arguments like `--tls-cert`, `--tls-key`, etc. to both dask-scheduler and dask-worker. This PR allows for arbitrary
additional command line arguments, so that any valid command line args to dask-scheduler and dask-worker are supported.

### DIFF
--- a/dask_cloudprovider/cloudprovider.yaml
+++ b/dask_cloudprovider/cloudprovider.yaml
@@ -4,9 +4,11 @@ cloudprovider:
     fargate_workers: False # Use fargate mode for the workers
     scheduler_cpu: 1024 # Millicpu (1024ths of a CPU core)
     scheduler_mem: 4096 # Memory in MB
+#   scheduler_extra_args: "--tls-cert,/path/to/cert.pem,--tls-key,/path/to/cert.key,--tls-ca-file,/path/to/ca.key"
     worker_cpu: 4096 # Millicpu (1024ths of a CPU core)
     worker_mem: 16384 # Memory in MB
     worker_gpu: 0 # Number of GPUs for each worker
+#   worker_extra_args: "--tls-cert,/path/to/cert.pem,--tls-key,/path/to/cert.key,--tls-ca-file,/path/to/ca.key"
     n_workers: 0 # Number of workers to start the cluster with
     scheduler_timeout: "5 minutes" # Length of inactivity to wait before closing the cluster
 
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -384,7 +384,8 @@ class Worker(Task):
                 "{}GB".format(int(self._mem / 1024)),
                 "--death-timeout",
                 "60",
-            ] + (list() if not extra_args else extra_args)
+            ]
+            + (list() if not extra_args else extra_args)
         }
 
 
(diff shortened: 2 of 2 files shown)


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -571,9 +592,12 @@ def __init__(
         self._scheduler_cpu = scheduler_cpu
         self._scheduler_mem = scheduler_mem
         self._scheduler_timeout = scheduler_timeout
+        self._scheduler_extra_args = scheduler_extra_args
         self._worker_cpu = worker_cpu
         self._worker_mem = worker_mem
         self._worker_gpu = worker_gpu
+        self._worker_extra_args = worker_extra_args
+        self._worker_extra_args = worker_extra_args
```
martindurant, line 600: This line appears twice

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -656,12 +680,18 @@ async def _start(self,):
         if self._scheduler_timeout is None:
             self._scheduler_timeout = self.config.get("scheduler_timeout")
 
+        if self._scheduler_extra_args is None:
+            self._scheduler_extra_args = self.config.get("scheduler_extra_args")
```
martindurant, line 684: Can you please add the two extra config options to the example config file, commented out and with a brief description

martindurant:
Thanks, this looks good.
I added a couple fo comments.

I wonder, how would you test this?

dask_cloudprovider/providers/aws/ecs.py
joeschmid: Sorry, thanks!

dask_cloudprovider/providers/aws/ecs.py
joeschmid: Done. Also, added splitting of comma-separated arguments if loaded from config file. (Arguments in existing code are in a list and I extended that list with any extra arguments.)

joeschmid:
> I wonder, how would you test this?

I've tested it successfully with Fargate & EC2, for configuring TLS encryption via arguments for --tls-cert, etc. Dask scheduler had `tls://` as the scheme for the URL of it's address, workers connected successfully using TLS, etc.

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -4,9 +4,11 @@ cloudprovider:
     fargate_workers: False # Use fargate mode for the workers
     scheduler_cpu: 1024 # Millicpu (1024ths of a CPU core)
     scheduler_mem: 4096 # Memory in MB
+#   scheduler_extra_args: "--tls-cert,/path/to/cert.pem,--tls-key,/path/to/cert.key,--tls-ca-file,/path/to/ca.key"
```
martindurant, line 7: Glad to see the example here.
Is there anything gained, do you think, by also supporting whitespace-delimited (for copy&paste from the command line) or lists (for copy&paste from code)?

joeschmid: I like the whitespace-delimited idea. That would definitely be convenient for copy/paste from actual command lines. Although for generic command line arguments (that could have quoted arguments with spaces) it could get tricky to parse, e.g.

worker_extra_args: "--name \"Joe Worker Type\""

Based on that I lean toward punting on support for whitespace delimiting, but let me know what you think. If you feel strongly about it, I don't mind taking a crack at parsing for it and see how it goes. (Have to be aware of quoted arguments and avoid splitting on those, etc.)

martindurant: If you support a list here, then the user can be explicit. I don't recommend trying to be clever here. Perhaps there is a similar situation elsewhere in the config? 
To be sure, I am OK to accept the PR as it is, until someone were to ask for something more comprehensive.

joeschmid: I did a quick review and all other config values are loaded directly, without any additional parsing, e.g. `self.image = self.config.get("image")`

I like the idea to wrap up this PR and then revisit if someone asks for something more comprehensive. Thanks!

martindurant:
Thanks, @joeschmid

PR #92, opened 2020-04-28
Support specifying platformVersion for Fargate

joeschmid:
Fargate allows for specifying the platformVersion with strings like "1.4.0" or "LATEST". Different platform versions have different capabilities, e.g. mounting EFS drives, etc. This PR allows users to specify a platform version string.

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -73,6 +73,10 @@ class Task:
         Configurable timeout in seconds for finding the task IP from the
         cloudwatch logs.
 
+    platform_version: str (optional)
+        Version of the AWS Fargate platform to use, e.g. "1.4.0" or "LATEST". This
+        setting has no effect for the EC2 launch type.
+
     kwargs:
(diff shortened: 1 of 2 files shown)


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -527,6 +535,11 @@ class ECSCluster(SpecCluster):
         and this operation takes a while.
 
         Default ``False``.
+    platform_version: str (optional)
+        Version of the AWS Fargate platform to use, e.g. "1.4.0" or "LATEST". This
```
martindurant, line 539: If None, this gets the value from the config - what happens if it is not set in the config, do you get LATEST?

joeschmid: It's ok if it's not set in the config and stays as None -- that would cause the code to skip L211 and just not set `platformVersion` in kwargs. That will cause AWS itself to will use LATEST. (AWS does that if you don't set a platformVersion. e.g. They have LATEST set to "1.3.0" right now, even though "1.4.0" is available.)

Issue #48, opened 2019-12-04
Add support for spot termination notices


jacobtomlinson:
Following on from discussions in dask/distributed#3300 it would be great if workers gracefull closed when they recieve a spot termination notification. This will be even more useful with #47.

- `ECSCluster` and `FargateCluster` workers should start with a `--preload 'dask_cloudprovider.aws.spot_termination_handler'` or similar. This could also be used for manual cluster creation separately from close cluster managers.
- This new `dask_cloudprovider.aws.spot_termination_handler` module should attach a task to the workers `periodic_callbacks` to poll the Spot Termination endpoint.
- When it gets a spot termination notice it should call `close_gracefully` on the worker.

jacobtomlinson:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-instance-termination-notices.html

Issue #47, opened 2019-12-04
Add support for Fargate Spot


jacobtomlinson:
> Fargate Spot is a new capability on [AWS Fargate](https://aws.amazon.com/fargate/) that can run interruption tolerant [Amazon Elastic Container Service (Amazon ECS)](https://aws.amazon.com/ecs/) Tasks at up to a 70% discount off the Fargate price.

https://aws.amazon.com/blogs/aws/aws-fargate-spot-now-generally-available/

Looks like AWS have introduced "capacity providers" which allow you to assign tasks to different capacity modes. Current modes are regular and spot.

Spot tasks can be interrupted at any time with a two minute warning in exchange for an up to 70% saving.

rsignell-usgs:
The blog post shows using the console, but it looks like using the API we need to first [create a capacity provider](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateCapacityProvider.html) and then [set `capacityProviderStrategy` to `FARGATE_SPOT`](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_CreateService.html#API_CreateService_RequestSyntax)  

@jacobtomlinson does that look about right?

jacobtomlinson:
That sounds about right to me.

JacksonMaxfield:
Just curious if there were any updates on this?

TomAugspurger:
I don't believe so. @jacobtomlinson (the primary dask-cloudprovider maintainer) is out for a couple months.

jthielen:
@tracek Would you happen to have your own implementation for fargate spot instance use mentioned in https://github.com/dask/dask-cloudprovider/pull/97#issuecomment-707161685 up somewhere or a PR ready? I thought I'd reach out to see what you may happen to already have before I try coming up with something myself (as I have need of this for a current project). Thanks!

tracek:
@jthielen Sorry for the lag, busy week. Yes, I do, and I am preparing a PR with this feature. There were some change in the meantime, so I need to figure out how to merge the changes properly. If you need something here and now, I can send you the package I built locally. Otherwise I hope to have something next week.

rpanai:
Hi @tracek do you need any help for this PR?

tracek:
[...]

PR #97, opened 2020-05-25
Support for Fargate Spot Capacity Provider

OverlordQ:
Per boto3 documentation:

> To use a AWS Fargate capacity provider, specify either the FARGATE or FARGATE_SPOT capacity providers. The AWS Fargate capacity providers are available to all accounts and only need to be associated with a cluster to be used.

There's no need to have to issue a separate `createCapacityProvider` request. 

Defaults to a 4:1 ratio for Spot/OnDemand capacity. 

Initial attempt at addressing Issue #47 

Does not handle spot termination notices per #48

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -867,22 +867,22 @@ class ECSCluster(SpecCluster):
             self._cluster_name_template
         )
         self.cluster_name = self.cluster_name.format(uuid=str(uuid.uuid4())[:10])
-        create_cluster_kwargs = {'clusterName': self.cluster_name, 'tags': dict_to_aws(self.tags)}
+        create_cluster_kwargs = {
+            "clusterName": self.cluster_name,
+            "tags": dict_to_aws(self.tags),
+        }
         if self._fargate_spot:
             spot_cluster_kwargs = {
-                'capacityProviders': ['FARGATE', 'FARGATE_SPOT'],
-                'defaultCapacityProviderStrategy': [
+                "capacityProviders": ["FARGATE", "FARGATE_SPOT"],
+                "defaultCapacityProviderStrategy": [
                     {
-                        'capacityProvider': 'FARGATE_SPOT',
-                        'weight': self._fargate_spot_weight
+                        "capacityProvider": "FARGATE_SPOT",
+                        "weight": self._fargate_spot_weight,
                     },
-                    {
-                        'capacityProvider': 'FARGATE',
-                        'weight': 1
-                    },
-                ]
+                    {"capacityProvider": "FARGATE", "weight": 1},
+                ],
             }
-            create_cluster_kwargs = { **create_cluster_kwargs, **spot_cluster_kwargs}
+            create_cluster_kwargs = {**create_cluster_kwargs, **spot_cluster_kwargs}
         async with self._client("ecs") as ecs:
             response = await ecs.create_cluster(**create_cluster_kwargs)
         weakref.finalize(self, self.sync, self._delete_cluster)
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -575,6 +575,8 @@ class ECSCluster(SpecCluster):
         self,
         fargate_scheduler=False,
         fargate_workers=False,
+		fargate_spot=False,
+		fargate_spot_weight=4,
         image=None,
         scheduler_cpu=None,
         scheduler_mem=None,
@@ -609,6 +611,8 @@ class ECSCluster(SpecCluster):
     ):
         self._fargate_scheduler = fargate_scheduler
         self._fargate_workers = fargate_workers
+		self._fargate_spot = fargate_spot
+		self._fargate_spot_weight = fargate_spot_weight
         self.image = image
         self._scheduler_cpu = scheduler_cpu
         self._scheduler_mem = scheduler_mem
@@ -672,6 +676,10 @@ class ECSCluster(SpecCluster):
             self._fargate_scheduler = self.config.get("fargate_scheduler")
         if self._fargate_workers is None:
             self._fargate_workers = self.config.get("fargate_workers")
+		if self._fargate_spot is None:
+		    self._fargate_spot = self.config.get("fargate_spot")
+		if self._fargate_spot_weight is None:
+		    self._fargate_spot_weight = self.config.get("fargate_spot_weight")
 
         if self._tags is None:
             self._tags = self.config.get("tags")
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -851,10 +867,24 @@ async def _create_cluster(self):
...
+                    },
+                    {"capacityProvider": "FARGATE", "weight": 1},
+                ],
+            }
+            create_cluster_kwargs = {**create_cluster_kwargs, **spot_cluster_kwargs}
         async with self._client("ecs") as ecs:
-            response = await ecs.create_cluster(
-                clusterName=self.cluster_name, tags=dict_to_aws(self.tags)
-            )
+            response = await ecs.create_cluster(**create_cluster_kwargs)
```
jacobtomlinson, line 887: Why have tags been removed here?

OverlordQ: Shouldn't be, line 852, tags and cluster name are part of the initial `create_cluster_kwargs`

jacobtomlinson: Ah great no problem!

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -851,10 +867,24 @@ async def _create_cluster(self):
...
+                "capacityProviders": ["FARGATE", "FARGATE_SPOT"],
+                "defaultCapacityProviderStrategy": [
+                    {
+                        "capacityProvider": "FARGATE_SPOT",
+                        "weight": self._fargate_spot_weight,
+                    },
+                    {"capacityProvider": "FARGATE", "weight": 1},
+                ],
+            }
+            create_cluster_kwargs = {**create_cluster_kwargs, **spot_cluster_kwargs}
```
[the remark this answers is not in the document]
OverlordQ: Done in latest commit

eganjs:
@OverlordQ I'm quite interested in making use of this feature, is there any chance of adding the necessary documentation and merging in the near future? I think many other people would be interested in using this too, it'd be a shame to let it just sit here when it's already code complete!

OverlordQ:
I'll see if I can dig up my notes and revise the PR

tracek:
@OverlordQ How to verify that indeed spot instances were used? Aside looking at Billing.

eganjs:
@tracek It's possibly a bit off topic, but you can look up those details with:
`ecs.describe_clusters(clusters=['<cluster arn>'], include=['STATISTICS'])`
See https://boto3.amazonaws.com/v1/documentation/api/1.9.42/reference/services/ecs.html#ECS.Client.describe_clusters for more info.

tracek:
@eganjs Thanks! I am asking this to verify if the PR works as expected, so in my mind this is very much on the topic.

As far as I can see, `describe_clusters` does not provide any info on the count (or presence) of spot instances. Here's how the response looks like on my end:


```
{'clusters': [{'clusterArn': 'arn:aws:ecs*',
   'clusterName': 'dask-*',
   'status': 'ACTIVE',
   'registeredContainerInstancesCount': 0,
   'runningTasksCount': 11,
   'pendingTasksCount': 0,
   'activeServicesCount': 0,
   'statistics': [{'name': 'runningEC2TasksCount', 'value': '0'},
    {'name': 'runningFargateTasksCount', 'value': '11'},
    {'name': 'pendingEC2TasksCount', 'value': '0'},
    {'name': 'pendingFargateTasksCount', 'value': '0'},
    {'name': 'activeEC2ServiceCount', 'value': '0'},
    {'name': 'activeFargateServiceCount', 'value': '0'},
    {'name': 'drainingEC2ServiceCount', 'value': '0'},
    {'name': 'drainingFargateServiceCount', 'value': '0'}],
   'tags': [],
   'settings': [{'name': 'containerInsights', 'value': 'disabled'}],
   'capacityProviders': [],
   'defaultCapacityProviderStrategy': []}],
 'failures': [],
 'ResponseMetadata': {'RequestId': *',
  'HTTPStatusCode': 200,
  'HTTPHeaders': {'x-amzn-requestid': '*',
   'content-type': 'application/x-amz-json-1.1',
   'content-length': '784',
   'date': 'Tue, 22 Sep 2020 15:11:45 GMT'},
  'RetryAttempts': 0}}
```

Or maybe I am missing something?

eganjs:
@tracek I think I originally misunderstood your question. I have had a deeper look around and unfortunately I have been unable to find anything to indicate it's possible to determine whether a task is running on Fargate Spot or just Fargate.

I believe it would be possible to prove spot instances are being created. This could be done by setting the size of the cluster to be greater than the Fargate On-Demand resource count, which as of today the default is 500 (see https://docs.aws.amazon.com/AmazonECS/latest/developerguide/service-quotas.html).

To be honest if the capacity providers are in place and we are able to see them, I would be happy. I think testing more than that would be testing AWS itself, which does feel out of scope.

But that said, I am happy to be proven or convinced otherwise.

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -575,6 +583,8 @@ def __init__(
         self,
         fargate_scheduler=False,
         fargate_workers=False,
+        fargate_spot=False,
```
tracek, line 586: Methinks it's a bug. In turn, the param value from the config file won't be loaded.

Either change `fargate_spot=False` to `fargate_spot=None` or change the logic altogether. Later in the code you do the following check:

```
if self._fargate_spot is None:
    self._fargate_spot = self.config.get("fargate_spot")
``` 

Since `fargate_spot` is defined as `False`, the param value from the config file won't be loaded. To reproduce:

```
# define FARGATE_SPOT in cloudprovider.yaml

cluster = FargateCluster()
print(cluster._fargate_spot)
```

tracek (changes requested):
IMO there's a small bug in the logic that prevents FARGATE_SPOT from being loaded from the config file.

tracek:
@eganjs @OverlordQ  I think I have found the root cause of why I could not find the right capacity provider - please have a look at the review. Maybe it makes to add test(s) that verify that config is loaded as expected? Users would be set for a surprise if they expected spot instances but got 100% normal ones. Unfortunately it can easily get unnoticed in the Billing and yet have heavy impact.

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -575,6 +583,8 @@ def __init__(
...
+        fargate_spot_weight=4,
```
tracek, line 587: Again, setting it to a number prevents from getting the value from the config file.

tracek:
I can confirm this does not work, no spot instances are created. I took a stab at this and added this functionality myself. Should I make a PR?

mrocklin:
A PR would be welcome

jacobtomlinson:
I think given the lack of activity here from the original author I'm going to close this out.

PR #108, opened 2020-06-29
Make fewer status requests and add more details in failure cases

### LINKED ISSUE #105: Review the orchestration code to be less chatty
run.status vs run.get_status() # fetches latest - don't use the latter multiple times in a conditional

also let's look into a better way to signal cluster standup than the metric name "scheduler", which can also be a bit more efficient as is by passing the metric name into the query

 def get_metrics(self, name=None, ...

### LINKED ISSUE #106: Add more details in failure case
https://github.com/dask/dask-cloudprovider/blob/a662fe2af211e702df2c9751c8a6f90f4feb4b51/dask_cloudprovider/providers/azure/azureml.py#L407

Use run.get_details() and the error property for a better customer experience. Check out how our SDK throws inside wait_for_completion()

FredFLi:
Closes #105 
Closes #106


dask_cloudprovider/providers/azure/azureml.py
```diff
@@ -394,17 +394,19 @@ async def __create_cluster(self):
 
         self.__print_message("Waiting for scheduler node's IP")
         while (
-            run.get_status() != "Canceled"
-            and run.get_status() != "Failed"
+            run.status != "Canceled"
```
akshaya-a, line 397: good change here, but now we need to add a status refresh after the sleep. Consider instead a construction like

```
status = run.get_status()
while status != A and status != B:
    time.sleep(..)
    status = run.get_status()

File: dask_cloudprovider/providers/azure/azureml.py [hunk 2]
```diff
@@ -394,17 +394,19 @@ async def __create_cluster(self):
...
             print(".", end="")
             logger.info("Scheduler not ready")
             time.sleep(5)
 
-        if run.get_status() == "Canceled" or run.get_status() == "Failed":
-            logger.exception("Failed to start the AzureML cluster")
-            raise Exception("Failed to start the AzureML cluster.")
+        if run.status == "Canceled" or run.status == "Failed":
+            run_details = run.get_details()
+            error_msg = run_details["error"]
```
akshaya-a, line 407: use .get() so you don't throw a KeyError instead

dask_cloudprovider/providers/azure/azureml.py
```diff
@@ -715,9 +717,9 @@ def update():
 
     def close_when_disconnect(self):
         if (
-            self.run.get_status() == "Canceled"
-            or self.run.get_status() == "Completed"
-            or self.run.get_status() == "Failed"
+            self.run.status == "Canceled"
```
akshaya-a, line 720: again here, i'd consider something like the below to get the latest status once as this call could be disjoint in time from the previous refresh

```
status = run.get_status()
if status == A or status == B....:

dask_cloudprovider/providers/azure/azureml.py
FredFLi: is the refresh needed? run.status show latest state ?

akshaya-a: to fetch the latest status, you need to use get_status(), which makes a network call

After making a network call, within the scope of a logical unit, you'd want to reuse the information you just got (like run.status or a temporary variable). run.status doesn't get you the latest status

FredFLi: updated

dask_cloudprovider/providers/azure/azureml.py
FredFLi: Good catch.

jacobtomlinson:
Looks like linting has failed. Could you please run `black` on the code.

jacobtomlinson:
Thanks!

PR #109, opened 2020-06-30
Add max retry for socket connection

FredFLi:
Cluster creation stuck on "on the same VNet" check if socket connection throws exception other than timeout and ConnectionRefusedError


dask_cloudprovider/providers/azure/azureml.py
```diff
@@ -353,12 +353,12 @@ async def __check_if_scheduler_ip_reachable(self):
             self.__print_message("On the same VNET")
             logger.info("On the same VNET")
         except socket.timeout as e:
-
             self.__print_message("Not on the same VNET")
-            logger.info("On the same VNET")
+            logger.info("Not on the same VNET")
             self.same_vnet = False
-        except ConnectionRefusedError as e:
+        except Exception as e:
```
jacobtomlinson, line 359: Why are we moving to such a broad exception here?

FredFLi: Thanks Jacob. In the beginning, I tried to keep ConnectionRefusedError and add a broad Exception but found that they are doing the same thing. So I combined them together. 

```
except ConnectionRefusedError as e:
    logger.info(e)
    self.__print_message(e)
    pass
except Exception as e:
    logger.info(e)
    self.__print_message(e)
    pass
```

jacobtomlinson: Sure I guess I'm asking what the need for the broad `Exception` is. We try to avoid them and only skip on expected exceptions, otherwise we may bury some unexpected issue.

FredFLi: That makes sense. I can revert it back. 
My thought is to log broad exceptions so that user knows why the check fails. For example, exception like the following:
`[Errno 113] No route to host`

jacobtomlinson: Sure that's reasonable. I just like to be explicit about these things.

Issue #111, opened 2020-07-14
Instantiating a FargateCluster when more than 50 log groups present fails if dask-ecs is > 50 in list of log groups


gdesmarais-ctx:
If you already have a log group named `dask-ecs` and it is not one of the first 50 log groups, the following call fails:

**What happened**:

A `ResourceAlreadyExistsException` exception is raised when instantiating a `FargateCluster`

**What you expected to happen**:

The `FargetCluster` instance is created

**Minimal Complete Verifiable Example**:

```python
from dask_cloudprovider.providers.aws.ecs import FargateCluster
fc = FargateCluster(vpc='vpc-nnnnnnn')
```

**Anything else we need to know?**:

The exception given is (first traceback is relevant one):

```
Traceback (most recent call last):
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/distributed/utils.py", line 323, in f
    result[0] = yield future
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/tornado/gen.py", line 735, in run
    value = future.result()
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/dask_cloudprovider/providers/aws/ecs.py", line 771, in _start
    or await self._create_cloudwatch_logs_group()
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/dask_cloudprovider/providers/aws/ecs.py", line 961, in _create_cloudwatch_logs_group
    await logs.create_log_group(logGroupName=log_group_name, tags=self.tags)
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/aiobotocore/client.py", line 134, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ResourceAlreadyExistsException: An error occurred (ResourceAlreadyExistsException) when calling the CreateLogGroup operation: The specified log group already exists
distributed.deploy.spec - WARNING - Cluster closed without starting up
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "/Users/gdesmarais/git/rightsize/rightsize-venv/lib/python3.7/site-packages/distributed/deploy/spec.py", line 621, in close_clusters
    cluster.close(timeout=10)
[...]

PR #113, opened 2020-07-14
basic cycling through paging for aws api

gregdes:
keep fetching until nextToken is empty
fixes https://github.com/dask/dask-cloudprovider/issues/111


gdesmarais-ctx:
> This looks great many thanks!
> 
> Happy for the log group name fix to be added here too.

The log group name fix would take more thinking upon inspection.  I'm not going to put it in this one - will create a new issue/branch after some thought.  Thanks @jacobtomlinson

jacobtomlinson:
Thanks @gdesmarais-ctx

Issue #116, opened 2020-07-30
AzureML Provider : Issues with resource group while creating AML Compute


arnabbiswas1:
While using the Azure ML Provider using the latest 0.4.0 package from [pypi](https://pypi.org/project/dask-cloudprovider/), I encountered with the following issues:

1. Not necessarily the resource group of the AML Workspace and the VNET would always be the same. If those are different, then compute cluster creation is failing while mounting the data store. The following code assumes that if VNET and SUBNET values are passed, then resource group will be same as the resource group of the AML workspace which is not correct.

```
        if self.vnet and self.subnet:
            vnet_rg = self.workspace.resource_group
            vnet_name = self.vnet
            subnet_name = self.subnet
```

IMHO following should do. If the resource_group is explicitly passed use it, otherwise, use the resource group associated with ML workspace.
```
	if self.vnet and self.subnet:
            vnet_name = self.vnet
            subnet_name = self.subnet
	    if self.resource_group:
	    	vnet_rg = self.resource_group
	    else:
	    	vnet_rg = self.workspace.resource_group
```

2. Initially I was creating the AML resource using the following code (using Jupyter Lab from Compute Instances):
```
		amlcluster = AzureMLCluster(ws,
                            vm_size="STANDARD_DS13_V2", 
                            datastores=[Datastore.get(ws, "my_datastore")], 
                            environment_definition=ws.environments['AzureML-Dask-CPU'], 
                            initial_node_count=1, 
                            scheduler_idle_timeout=600)
```
This was failing while mounting the datastore with the following error message (since I didn't configure VNET as mentioned in #1):

> ERROR - Failed to start the AzureML cluster. {'error': {'code': 'UserError', 'message': 'AzureMLCompute job failed.\nBFSMountError: unable to mount blob fuse file system\n\tInfo: Could not mount Azure Blob Container some_name at my_datastore: Unauthorized. Cannot access the storage account with the given account key. Please verify that the account key is valid.\n\tInfo: Job environment preparation failed on 10.0.0.5.', 'details': []}, 'correlation': {'operation': None, 'request': 'aaaa'}, 'environment': 'aaa', 'location': 'aaa', 'time': '2020-07-30T00:00:00.00000'}
> NoneType: None

But by the time it fails, the physical compute cluster with 1 node has already been created. Since it failed, amlcluster object was not created and as a result I am not able to clean the compute cluster using:

```
    amlcluster.close()
````

As expected, resulting error message: 

[...]

PR #123, opened 2020-08-21
Clean up compute target in case the cluster creation fails

FredFLi:
1. Clean up compute target if creation fails. #116  2nd task
2. Fix timeout due to private key file permission.
3. Check ssh key and admin username if user provides compute target


lostmygithubaccount:
@jacobtomlinson can you review this? fixes some major issues in the released version

jacobtomlinson (approved):
This is great many thanks. Will get out a bug fix release with this in.

jacobtomlinson:
This has been included in `0.4.1`

PR #124, opened 2020-08-25
Backoff and retry status polling

### LINKED ISSUE #56: "Rate exceeded" error on FargateCluster
```
from dask_cloudprovider import FargateCluster
cluster = FargateCluster(n_workers=1, image='rsignell/pangeo-worker:2020-01-23c')
client = Client(cluster)
```
worked on my SageMaker instance on uswest-2, but failed my useast-1 instance with:
```python-traceback
Task exception was never retrieved
future: <Task finished coro=<_wrap_awaitable() done, defined at /home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/asyncio/tasks.py:530> exception=ClientError('An error occurred (ThrottlingException) when calling the DescribeTasks operation (reached max retries: 4): Rate exceeded',)>
Traceback (most recent call last):
  File "/home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/asyncio/tasks.py", line 537, in _wrap_awaitable
    return (yield from awaitable.__await__())
  File "/home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/site-packages/dask_cloudprovider/providers/aws/ecs.py", line 130, in _
    await self.start()
  File "/home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/site-packages/dask_cloudprovider/providers/aws/ecs.py", line 240, in start
    await self._update_task()
  File "/home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/site-packages/dask_cloudprovider/providers/aws/ecs.py", line 158, in _update_task
    cluster=self.cluster_arn, tasks=[self.task_arn]
  File "/home/ec2-user/SageMaker/myenvs/pangeo/lib/python3.6/site-packages/aiobotocore/client.py", line 102, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (ThrottlingException) when calling the DescribeTasks operation (reached max retries: 4): Rate exceeded
```
It looks like the PR https://github.com/dask/dask-cloudprovider/pull/44 was designed to address at these problems, but it seems I'm still having them despite running v0.1.1 which includes that PR.  I'm wondering whether others are still experiencing this?

jacobtomlinson:
Fixes #56.

We currently poll for the pending container status every second until it is running. If ECS takes a long time to start we can eventually run into a `ThrottlingException`.

This PR catches the exception and continues retrying but with an exponential backoff to a maximum of 20 seconds. Once a request is successful it will resume polling every second until the exception is hit again.

This will continue indefinitely until the container moves to a running or failed state.

### DIFF
--- a/dask_cloudprovider/providers/aws/ecs.py
+++ b/dask_cloudprovider/providers/aws/ecs.py
@@ -247,9 +247,17 @@ class Task:
                 await asyncio.sleep(1)
 
         self.task_arn = self.task["taskArn"]
+        wait_duration = 1
         while self.task["lastStatus"] in ["PENDING", "PROVISIONING"]:
-            await asyncio.sleep(1)
-            await self._update_task()
+            try:
+                await self._update_task()
+                wait_duration = 1
+            except ClientError as e:
+                if e.response["Error"]["Code"] == "ThrottlingException":
+                    wait_duration = wait_duration * 2
+                else:
+                    raise
+            await asyncio.sleep(min(wait_duration, 20))
         if not await self._task_is_running():
             raise RuntimeError("%s failed to start" % type(self).__name__)
         [eni] = [


eganjs (changes requested):
Looks pretty great, just a few comments:
- Can the maximum backoff can be made configurable?
- Should a note be added in the documentation to cover this?

jacobtomlinson:
Thanks for the review @eganjs.

The 20 second maximum is described in the AWS documentation. Do you have a reason why you would like this to be configurable?

I'm not sure this needs to be documented, this is more of a bug fix to respect the AWS API. It shouldn't affect the usage of the library.

eganjs (approved):
That's a good challenge.

I've re-read through #56 and the cases I was thinking of would work fine under the implementation.

I think I was being a bit overzealous with my concerns 😅

eganjs:
Today the default limits for `Fargate On-Demand resource count` and `Fargate Spot resource count` were increased from 100 to 500 and 250 to 500 respectively (see https://aws.amazon.com/about-aws/whats-new/2020/09/aws-fargate-increases-default-resource-count-service-quotas/). Without this PR I suspect users of the library will be unable to leverage these increases, it would be good to get this merged ASAP unless there is more work to be done here?

jacobtomlinson:
Thanks for the nudge @eganjs

PR #125, opened 2020-08-26
Extract provider specific requirements into extras_require in setup.py

eganjs:
Hi there, I've raised this PR to attempt to improve the developer experience for people only interested in one cloud provider!

### DIFF
--- a/dask_cloudprovider/__init__.py
+++ b/dask_cloudprovider/__init__.py
@@ -1,6 +1,13 @@
 from . import config
-from .providers.aws.ecs import ECSCluster, FargateCluster
-from .providers.azure.azureml import AzureMLCluster
+
+try:
+    from .providers.aws.ecs import ECSCluster, FargateCluster
+except ImportError:
+    pass
+try:
+    from .providers.azure.azureml import AzureMLCluster
+except ImportError:
+    pass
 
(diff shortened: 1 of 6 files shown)


jacobtomlinson (changes requested):
This is awesome thanks. Funnily enough I was thinking about the same thing this week.

Could I request a small change. It would be great to also have an `all` option to get everything. You could do something like this.

```python
# setup.py

extras_require = {
    "aws": ["aiobotocore>=0.10.2"],
    "azure": ["azureml-sdk>=1.0.83"],
}
extras_require["all"] = set(
    package for packages in extras_require.values() for package in packages
)

...

setup(
    ...
    extras_require=extras_require,
    ...
)
```

Then in the docs you can put something like this

```rst
Pip
^^^

.. code-block:: console

   $ pip install dask-cloudprovider[all]

You can also restrict your install to just a specific cloud provider by giving their name instead of ``all``.

.. code-block:: console

   $ pip install dask-cloudprovider[aws]  # or
   $ pip install dask-cloudprovider[azure] 
```

eganjs:
@jacobtomlinson Agreed and done. The `all` extra is a nice addition.

jacobtomlinson (approved):
Awesome thanks for doing this so quickly!

PR #129, opened 2020-09-04
Add VMCluster support

jacobtomlinson:
So far all of the existing cluster managers in `dask-cloudprovider` use high level services. This PR adds a generic base class called `VMCluster` which is designed to create a Dask cluster comprised of VMs.

The base class is intended to be inhertied from with cloud provider specific versions and this PR also includes a DigitalOcean reference implementation.

The `VMCluster`, `VMScheduler` and `VMWorker` base classes have the following assumptions.

- Schedulers and workers will each be a whole VM
- VMs will be running Ubuntu and configured via cloud-init/user data
- Schedulers will always run remotely


mrocklin:
Ooh, shiny

dask_cloudprovider/providers/digitalocean/tests/test_droplet.py
```diff
@@ -0,0 +1,35 @@
+import pytest
+
+from dask_cloudprovider.providers.digitalocean.droplet import DropletCluster
+from dask.distributed import Client
+import dask.array as da
+from distributed.core import Status
+
+
+@pytest.fixture
+async def gen_cluster():
```
mrocklin, line 10: I recommend not using the `gen_cluster` name here.  Maybe instead just `cluster`?

gen_cluster came from `tornado.gen` which probably doesn't make as much sense anymore :)

dask_cloudprovider/providers/digitalocean/tests/test_droplet.py
```diff
@@ -0,0 +1,35 @@
...
+    async with DropletCluster(asynchronous=True) as cluster:
+        yield cluster
+
+
+@pytest.mark.asyncio
+async def test_init():
+    cluster = DropletCluster(asynchronous=True)
+    assert cluster.status == Status.created
```
mrocklin, line 18: `async with`?

dask_cloudprovider/providers/digitalocean/tests/test_droplet.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.timeout(300)
+# @pytest.mark.integration
+async def test_create_cluster(gen_cluster):
+    cluster = await gen_cluster
```
mrocklin, line 25: Is this not automatic from the `async with` in `gen_cluster`?

dask_cloudprovider/providers/digitalocean/tests/test_droplet.py
```diff
@@ -0,0 +1,35 @@
...
+    assert cluster.status == Status.running
+
+    cluster.scale(2)
+
+    await cluster
+
+    assert len(cluster.workers) == 2
+
+    client = Client(cluster, asynchronous=True)  # noqa
+    await client
```
mrocklin, line 35: You might consider including a fixture that makes sure that things get cleaned up.  We have `distributed.utils_test.cleanup` in that package, which would err here because you have a lingering `Client`.  That might be overkill though.

dask_cloudprovider/providers/generic/vmcluster.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    def __init__(self, **kwargs):
+        self.docker_image = "daskdev/dask:latest"
+        super().__init__(**kwargs)
+
+    def render_cloud_init(self, *args, **kwargs):
+        loader = FileSystemLoader([os.path.dirname(os.path.abspath(__file__))])
+        environment = Environment(loader=loader)
+        template = environment.get_template("cloud-init.yaml.j2")
```
mrocklin, line 22: I'm curious, why the `.j2` suffix?

dask_cloudprovider/providers/generic/vmcluster.py
```diff
@@ -0,0 +1,102 @@
...
+        return template.render(*args, **kwargs)
+
+    async def start(self):
+        """Create a VM."""
+        await super().start()
+
+    async def close(self):
+        """Destroy a VM."""
+        await super().close()
```
mrocklin, line 31: These are here to be suggestive to downstream authors?

mrocklin (commented):
Some small comments, none of them very inspired :)

dask_cloudprovider/providers/digitalocean/tests/test_droplet.py
jacobtomlinson: I am intentionally not awaiting the cluster here

dask_cloudprovider/providers/generic/vmcluster.py
jacobtomlinson: Jinja2 :)

dask_cloudprovider/providers/generic/vmcluster.py
jacobtomlinson: Yeah.

dask_cloudprovider/providers/generic/vmcluster.py
```diff
@@ -0,0 +1,135 @@
...
+    def render_cloud_init(self, *args, **kwargs):
+        loader = FileSystemLoader([os.path.dirname(os.path.abspath(__file__))])
+        environment = Environment(loader=loader)
+        template = environment.get_template("cloud-init.yaml.j2")
+        return template.render(**kwargs)
+
+    @classmethod
+    def get_cloud_init(cls, *args, **kwargs):
+        cluster = cls(*args, asynchronous=True, **kwargs)
+        cluster.auto_shutdown = False
```
quasiben, line 128: We don't want this hardcoded here, correct ?

jacobtomlinson: I think for the Packer use case it makes sense to default this to `False`, but perhaps now that it is configurable it should be configurable here too.

doc/Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
+SPHINXOPTS    ?= -a
```
quasiben, line 6: What does `-a` do ?

doc/source/aws.rst
```diff
@@ -0,0 +1,41 @@
+Amazon Web Services (AWS)
```
quasiben, line 1: I like the reorganization from general API to cloud specific  listings

dask_cloudprovider/providers/aws/ecs.py
```diff
@@ -1214,14 +1197,126 @@ class FargateCluster(ECSCluster):
     kwargs: dict
         Keyword arguments to be passed to :class:`ECSCluster`.
 
+    Examples
```
quasiben, line 1200: The intention here is that this will be rendered in RTD, correct ?

doc/source/index.rst
```diff
@@ -3,268 +3,74 @@ Dask Cloud Provider
...
-                   "*"
-               ]
-           }
-       ],
-       "Version": "2012-10-17"
-   } 
+   Cluster managers will attempt to automatically remove hanging cloud resources on garbage collection if the cluster
+   object is destroyed without calling ``cluster.close()``, however this is not guaranteed.
 
-GPU Support
```
quasiben, line 176: Maybe we should keep some information around about launching GPU instances and link to the specific examples like ecscluster+rapids ?

doc/Makefile
jacobtomlinson: It rewrites all files even if nothing changes. I was having some caching issues where the sidebar wasn't updating when adding to the TOC. Happy to take this out again.

dask_cloudprovider/providers/aws/ecs.py
jacobtomlinson: Yeah

doc/source/index.rst
jacobtomlinson: This ECS example was moved into the docstring of `ECSCluster`. We could add a general GPU section to the index though.

PR #144, opened 2020-10-11
Updates to AzureMLCluster

lostmygithubaccount:
- `environment_definition` -> `environment` (breaking API change, probably not allowed?)
- remove datastores
- `Estimator` -> `ScriptRunConfig` and necessary changes to inputs, removing dependency on azureml.train


jacobtomlinson:
Thanks for this. I'm not too worried about the breaking change at this point. This API is still a little in flux.

nickeubank:
Any chance this addresses #150 ?

lostmygithubaccount:
don't think so - commented over there - I also think this PR needs:

- [ ] testing
- [ ] update documentation
- [ ] other edits?

lostmygithubaccount:
@jacobtomlinson would it be possible to setup an Azure ML Workspace so we could do some automated testing with GitHub actions?

jacobtomlinson:
@lostmygithubaccount I've been thinking about better testing today actually.

The problem is that we would need to store credentials in GitHub secrets which are not available in PRs (and for good reason). Do you have thoughts on a solution to this? I'm looking into some kind of "ok to test" workflow now but not finding much.

lostmygithubaccount:
we have the same problem in Azure/azureml-examples - it does work on PRs for *branches*, just not *forks* - so if you're comfortable giving contributors the ability to create branches that could help

jacobtomlinson:
That could work! I'll look into it.

lostmygithubaccount:
close in favor of #154

jacobtomlinson:
Do you favour generic Azure VM support over AzureML all together?

lostmygithubaccount:
@jacobtomlinson yes, should be more stable, maintainable, and hopefully faster

PR #147, opened 2020-10-20
Add local scheduler support

skozlovf:
In most cases ECS cluster uses the same instance type for containers. And if you need powerful instance with single worker inside then scheduler will run in the same powerful instance which is not optimal. For example:
```
ECS EC2 cluster:
m5.16xlarge - worker
m5.16xlarge - worker
m5.16xlarge - scheduler
```

Options here are:

* Run the scheduler locally if client runs in a EC2 instance.
* Use separate instance type for the scheduler, but this may require to set `placementConstraints` for the scheduler task. `ECSCluster` doesn't support extra task arguments, so it's not possible yet.

The PR implements first way.

Usage:
```python
 with ECSCluster(local_scheduler=True, ...) as cluster:
    ...

 with ECSCluster(local_scheduler={"options": {"port": 9090, "dashboard_address": ":9099"}}, ...) as cluster:
    ...

 with ECSCluster(local_scheduler={"options": {"port": 9090, "dashboard_address": ":9099"}, "cls": MyScheduler}, ...) as cluster:
    ...
```


jacobtomlinson (commented):
Thanks for this. I agree this is a useful thing to add.

My only comment here is that in dask-kubernetes this kwarg is called `deploy_mode`.

https://github.com/dask/dask-kubernetes/blob/ccb9864d463a3e14988af53a947a658cf32e7e10/dask_kubernetes/core.py#L301-L303

It would be good to keep things consistent here. Although I think it makes sense for `"remote"` to be the default in this case.

skozlovf:
Didn't note it's already pupported in the `dask-kubernetes`.
Not sure what to change.  Rename `local_scheduler` to `deploy_mode` and process similar to `dask-kubernetes`?

jacobtomlinson:
>  Rename local_scheduler to deploy_mode and process similar to dask-kubernetes?

Yeah that would be great. But dask-kubernetes has `local` as the default, but in this case I think `remote` should be the default.

skozlovf:
Updated, looks more consistent now, but I left possibility to specify local cluster spec. Take a look please.

dask_cloudprovider/aws/ecs.py
```diff
@@ -615,6 +616,14 @@ class ECSCluster(SpecCluster):
         mounted in worker tasks. This setting controls whether volumes are also mounted in the scheduler task.
 
         Default ``False``.
+    local_scheduler_spec: dict (optional)
+        Local scheduler specification. If provided, then ``deploy_mode`` auto sets to ``"local"``.
+
+        Default ``None``.
```
jacobtomlinson, line 622: I'm curious what this is used for. Could you share why you would pass an alternate local scheduler spec?

skozlovf: It can be used if run cluster outside AWS. For example, if you use NAT, from home PC/laptop behind WiFi router.
In my case I need to set specific port, router is configured to forward Inet connection to the port.

```                     
[worker] --> [Public IP, router:port1]  ---> [private network, XXX:port2]
```

```python

    with ECSCluster(
            ...,
            deploy_mode="local",
            local_scheduler_spec={
                "options": {
                    "host": "192.168.1.52",   # XXX
                    "port": 8989                    # port2
                }
            },
    ) as cluster:
```

In short, it allows to pass options to scheduler or(and) own implementation class.
 `dask-kubernetes` and `dask-cloudprovider` allow to run only default scheduler.

If it's too specific case, then I'm ok to drop it.

jacobtomlinson: > If it's too specific case, then I'm ok to drop it.

It may be too specific.

To be honest these exact cases are the reason why the scheduler is remote in the first place. Having the workers talk to a scheduler over the internet is likely going to be a worse experience than if the scheduler was inside AWS too. You're likely introducing unnecessary latency here.

I'm also a little confused because the scheduler should not be placed on its own instance. If this is the case there may be a problem with the way containers are being packed. The scheduler container is smaller than the workers, but if your workers fill the whole instance then there will be no room for the scheduler anywhere and it will fit onto its own machine.

I guess I always imagined that instances in an ECS cluster would be pretty large, and workers, schedulers and other workloads would all be run together. So the scheduler would handle efficient packing.

It actually sounds like you are using an ECS cluster just to run one Dask clusters. If this is the case it may be simpler to use `EC2Cluster` instead. The same problems exist there, but the solutions may fit a little better (placing the scheduler on a smaller instance type than the workers).

skozlovf: > I'm also a little confused because the scheduler should not be placed on its own instance. If this is the case there may be a problem with the way containers are being packed. The scheduler container is smaller than the workers, but if your workers fill the whole instance then there will be no room for the scheduler anywhere and it will fit onto its own machine.

It's not placed on its own instance, it just runs in the `Client` instance. Mostly to to save resources in some cases.

To solve scheduler packing issue I proposed to support extra tasks parameters (already merged), this way we can run less powerful instance for the scheduler container.

As for `EC2Cluster`, you are right. I came to this solution. But for different reason, I need more flexible EBS volumes support. Thanks for the suggestion.

Overall make sense to me and I got your point. Will update code soon.

jacobtomlinson: Thanks for the updates.

You may also be interested in dask/distributed#4263. Which would make a local scheduler much less desirable.

skozlovf: Thanks, will check the issue.

jacobtomlinson:
I think given the direction I want to nudge things in dask/distributed#4263 I'm going to close this out. Adding a local scheduler makes things generally less flexible.

PR #156, opened 2020-11-05
[docs] document valid config options for ECS / Fargate

jameslamb:
I've been using `dask-cloudprovider` a lot recently to test things and come up with reproducible examples.

Through this experience, I learned that ECS / Fargate only allow task definitions to choose from a finite list of CPU and memory settings. This is an AWS limitation and not a `dask-cloudprovider` one, but it took me a bit of investigation to figure that out.

> botocore.errorfactory.ClientException: An error occurred (ClientException) when calling the RegisterTaskDefinition operation: No Fargate configuration exists for given values.

This PR proposes adding a link to the [valid config values for ECS](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html) in the docs, so that when people search the `dask-cloudprovider` docs or Google for this error, they quickly find the answer and the set of valid values.

Thanks for your time and consideration.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -452,11 +452,13 @@ class ECSCluster(SpecCluster):
     scheduler_cpu: int (optional)
         The amount of CPU to request for the scheduler in milli-cpu (1/1024).
 
-        Defaults to ``1024`` (one vCPU).
+        Defaults to ``1024`` (one vCPU). See
+        `the troubleshooting guide <./troubleshooting.html#invalid-cpu-or-memory>`_ for information on the valid values for this argument.
     scheduler_mem: int (optional)
         The amount of memory to request for the scheduler in MB.
 
-        Defaults to ``4096`` (4GB).
+        Defaults to ``4096`` (4GB). See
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
This is great thanks @jameslamb.

Should we also mention this (or link to the troubleshooting guide) from the docstrings that describe these kwargs.

https://github.com/dask/dask-cloudprovider/blob/9f3fc1d53ce263d8b75f2a747a72b6dabcd88eb4/dask_cloudprovider/aws/ecs.py#L452-L459

https://github.com/dask/dask-cloudprovider/blob/9f3fc1d53ce263d8b75f2a747a72b6dabcd88eb4/dask_cloudprovider/aws/ecs.py#L470-L477

jameslamb:
oh sure, good idea! I'll link to the troubleshooting guide from there, to avoid duplicating that AWS link into multiple places.

jameslamb:
Alright I just pushed an update, let me know what you think! I regenerated the docs like this:

```shell
pip install .[digitalocean]
cd doc
make html
```

Locally, after that change the link looks like this:

image

And I clicked through and confirmed that the relevant reference like `/troubleshooting.html#invalid-cpu-or-memory` works as expected.

jacobtomlinson:
Perfect thanks @jameslamb

PR #162, opened 2020-11-11
Fix GPU flag for A100

jacobtomlinson:
To use NVIDIA A100 GPUs on GCP you use either the `a2-highgpu` or `a2-megagpu` machine types instead of adding an accelerator.

For older GPUs including V100 and T4 you still specify an accelerator and the number of them to attach.

This PR ensured the cloud init is passed a correct `gpu_instance` regardless of the GPU selection method.

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -168,12 +168,16 @@ class GCPInstance(VMInterface):
 
         return config
 
+    @property
+    def gpu_instance(self):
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
I went with `"gpu" in self.machine_type` instead of `self.machine_type.startswith("a2")` in an attempt to be future proofed. But happy with either approach.

quasiben:
I think `gpu` in machine_type is good.  Thanks @jacobtomlinson !

jacobtomlinson:
Ok great. Merging!

Issue #161, opened 2020-11-11
GCP cluster connection timeout over external IP


eric-czech:
Very excited to see https://github.com/dask/dask-cloudprovider/pull/131 in!  I gave this a try again today @quasiben but had an issue.  I ran this:

```python
from dask_cloudprovider.gcp.instances import GCPCluster
cluster = GCPCluster(
    name='dask-gcp-test-1', 
    zone='us-east1-c', 
    machine_type='n1-standard-8', 
    projectid=MY_PROJECT_ID,
    docker_image="daskdev/dask:latest",
    ngpus=0
)
Launching cluster with the following configuration: 
  Source Image: projects/ubuntu-os-cloud/global/images/ubuntu-minimal-1804-bionic-v20201014 
  Docker Image: daskdev/dask:latest 
  Machine Type: n1-standard-8 
  Filesytsem Size: 50 
  N-GPU Type:  
  Zone: us-east1-c 
Creating scheduler instance
dask-7c88f69a-scheduler
	Internal IP: 10.142.0.2
	External IP: 34.75.186.200
Waiting for scheduler to run
# No connection made after ~10 mins so I killed the process
```

The scheduler connection never occurred despite this being run from a VM in GCP.  That would make sense if it's trying to connect with the external IP and I found that if I try to connect directly that way (via dask.distributed.Client), this also doesn't work.  I have no firewall rules set up to allow ingress to 8786 and I'd rather not add them.  Is there a way to have `GCPCluster` or its parent classes use internal IPs instead?  I would much prefer that.  If I try to connect using the internal IP in this case (again via dask.distributed.Client), everything is fine so I know the VM is up and running correctly.

How did you get around this in your testing?  Did you configure firewall rule excepts for the dask scheduler port or perhaps already have them in place?

quasiben:
Thanks @eric-czech for testing this out. Yes, we did test with firewall rules opening `tcp:8888,8787,22,8786`.  We were testing from outside of GCP a needed to ping the scheduler externally.  Currently, we capture both the internal and external IPs:

https://github.com/dask/dask-cloudprovider/blob/09c28979fd283d483f46ffec7e5165d887a9083a/dask_cloudprovider/gcp/instances.py#L263-L265

Then use the external address for assessing if the scheduler is up.  This could be configurable.  You are connecting to the scheduler from within GCP, correct ?  That is, your client can ping the internal IP address ?

eric-czech:
[...]

PR #164, opened 2020-11-11
Configure scheduler internal/external ip with public_ingress option

quasiben:
Resolves https://github.com/dask/dask-cloudprovider/issues/161

cc @eric-czech

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup Conda Environment
-        uses: goanpeca/setup-miniconda@v1.7.0
+        uses: conda-incubator/setup-miniconda@v2
         with:
           miniconda-version: "latest"
           python-version: "3.7"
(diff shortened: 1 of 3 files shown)


eric-czech:
🙏  @quasiben

quasiben:
@jacobtomlinson was able to launch a cluster in 3 minutes after generating a machine image with packer and baking in the docker image to the machine image.  My guess is that he'll have [additional instructions](https://cloudprovider.dask.org/en/latest/packer.html) up for this later this week

jacobtomlinson:
Should we also be setting some config somewhere for `GCPCluster` to not request a public IP address.

In here perhaps?

https://github.com/dask/dask-cloudprovider/blob/fcedd96ed173ba5c1a77fa35d6ac48787e44fb2e/dask_cloudprovider/gcp/instances.py#L108-L122

quasiben:
> Should we also be setting some config somewhere for GCPCluster to not request a public IP address.
In here perhaps?

Yes -- addressed in 8dd3789

PR #172, opened 2020-11-16
Fix bootstrapping on GCP

jacobtomlinson:
Due to a typo and missing attribute bootstrapping is being skipped on GCP. 

This appears to have been missed because the default GCP Ubuntu image comes with Docker pre installed.


quasiben:
👍

PR #174, opened 2020-11-18
Increase file descriptor limits on ECS/Fargate workers

lmarsden:
* This change resolves the issue of the maximum number of open file descriptors/network connections being exceeded at runtime. This is documented  in the Dask Distributed FAQs (https://distributed.dask.org/en/latest/faq.html). The issue occurs on `FargateCluster` nodes under some types of heavy workloads with large numbers of workers.
* The `nofile` hard and soft ulimits are set to `65535` (from default `1024`) in both the scheduler and worker nodes via the ECS task  definitions for each.
* Also see https://github.com/dask/distributed/issues/1858

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -1105,6 +1105,13 @@ class ECSCluster(SpecCluster):
                             if not self._scheduler_extra_args
                             else self._scheduler_extra_args
                         ),
+                        "ulimits": [
+                            {
+                                "name": "nofile",
+                                "softLimit": 65535,
+                                "hardLimit": 65535,
+                            },
+                        ],
                         "logConfiguration": {
                             "logDriver": "awslogs",
                             "options": {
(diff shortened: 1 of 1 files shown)


jacobtomlinson:
Awesome many thanks!

> The issue occurs on FargateCluster nodes under some types of heavy workloads with large numbers of workers.

I would love to hear more about how you are using dask-cloudprovider if you're in a position to share?

lmarsden:
You're very welcome!

We're using dask-cloudprovider to process inference workloads on large oceanographic datasets. In the situation that elicited the original issue we're using a `FargateCluster` with `worker_cpu=4096` but `--nthreads` set to `1` via `worker_extra_args`. This trick allows the inference processing on each worker to make use of all four vCPUs for multithreading, but without interfering with Dask's own multithreading, as the number of tasks per worker is restricted to one. It works well, but was falling over for larger cluster sizes (~100 nodes) with the file descriptor limit exceeded error. Sorted now, thanks for the swift merge.

PR #175, opened 2020-11-18
Add AzureVMCluster

### LINKED ISSUE #154: Generic Azure support
With the new VMCluster and GCP support being added, what would it take to add a generic Azure cloudprovider package?

jacobtomlinson:
Added a VM cluster implementation for Azure.

```python
>>> from dask_cloudprovider.azure import AzureVMCluster
>>> cluster = AzureVMCluster(resource_group="<resource group>",
...                          vnet="<vnet>",
...                          security_group="<security group>",
...                          n_workers=1)
Creating scheduler instance
Assigned public IP
Network interface ready
Creating VM
Created VM dask-5648cc8b-scheduler
Waiting for scheduler to run
Scheduler is running
Creating worker instance
Network interface ready
Creating VM
Created VM dask-5648cc8b-worker-e1ebfc0e
```

Closes #154

cc @lostmygithubaccount


setup.py
```diff
@@ -7,7 +7,12 @@
 
 extras_require = {
     "aws": ["aiobotocore>=0.10.2"],
-    "azure": ["azureml-sdk>=1.0.83"],
+    "azure": [
+        "azureml-sdk>=1.0.83",
```
lostmygithubaccount, line 11: can we add a new one called `azureml` or `aml` to separate out the `azureml-sdk` install? it's really bloated

jacobtomlinson: Yeah that might be a good idea.

jacobtomlinson: I think to do this neatly we would also need to move the azureml code into its own provider. So it would go from `dask_cloudprovider.azure.AzureMLCluster` to `dask_cloudprovider.azureml.AzureMLCluster`. But `dask_cloudprovider.azure.AzureVMCluster` would remain the same.

lostmygithubaccount: I'm going to make a PR to remove `AzureMLCluster` after this is in - maybe we should do it as one PR?

lostmygithubaccount: maybe we should leave it in with deprecation notice

jacobtomlinson: > I'm going to make a PR to remove AzureMLCluster after this is in

😂 let's deprecate it for a release or two before removing

jacobtomlinson:
@lostmygithubaccount given your feedback I've added the following changes:
- We now show a deprecation warning for `AzureMLCluster` when someone uses it
- It has been moved to `dask_cloudprovider.azureml.AzureMLCluster`
- Its dependencies are a separate extra install `pip install dask_cloudprovider[azureml]`

jacobtomlinson:
I think this is ready now. @lostmygithubaccount @quasiben would you be able to have a look?

dask_cloudprovider/azure/tests/test_azurevm.py
```diff
@@ -0,0 +1,114 @@
...
+
+@pytest.mark.asyncio
+@pytest.mark.timeout(1200)
+@skip_without_credentials
+async def test_create_rapids_cluster_sync():
+
+    with AzureVMCluster(
+        vm_size="Standard_NC12s_v3",
+        docker_image="rapidsai/rapidsai:cuda11.0-runtime-ubuntu18.04-py3.8",
+        worker_class="dask_cuda.CUDAWorker",
```
quasiben, line 98: In the future are we still planning on moving to worker_modules ?

doc/source/azureml.rst
```diff
@@ -0,0 +1,25 @@
...
+    The Azure ML integration has been deprecated and will be removed in a future release.
+    Please use the :class:`dask_cloudprovider.azure.AzureVMCluster` cluster manager instead.
+
+.. autosummary::
+   AzureMLCluster
+
+Overview
+--------
+
+In order to start using ``dask_cloudprovider.AzureMLCluster`` you need, at a minimum,
```
quasiben, line 17: Small nit:

```suggestion
To start using ``dask_cloudprovider.AzureMLCluster`` you need, at a minimum,
```

dask_cloudprovider/azure/tests/test_azurevm.py
quasiben: Oh, I see, you are handling both now.  Apologies for the noise and that's great!

quasiben:
Thanks @jacobtomlinson -- looks good!

dask_cloudprovider/azure/tests/test_azurevm.py
jacobtomlinson: Yeah handling both for now. Although I want to move everything over to the spec approach eventually.

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -48,19 +48,36 @@ cloudprovider:
...
-    scheduler_port: 9002 # default for the scheduler port if using from outside of cluster VNET
-    scheduler_idle_timeout: 1200 # default timeout for dask-scheduler process to shutdown if idle (in seconds)
-    worker_death_timeout: 30 # default timeout for dask-worker process upon losing scheduler
-    additional_ports: [] # list of tuples of additional ports to map/forward
-    admin_username: "" # username to log in to the AzureML Training Cluster for 'local' runs
-    admin_ssh_key: "" # password to log in to the AzureML Training Cluster for 'local' runs
-    telemetry_opt_out: False # by default we log the version of the AzureMLCluster being used, set to True to opt-out
-    datastores: [] # default list of datastores to mount
+    location: "eastus2" # The Azure location to launch your cluster
+    resource_group: null # The Azure resource group for the cluster
```
lostmygithubaccount, line 52: what will default resource group name be?

lostmygithubaccount (approved):

lostmygithubaccount:
looks good to me, thanks for this @jacobtomlinson!

dask_cloudprovider/cloudprovider.yaml
jacobtomlinson: It defaults to `None` which will raise a `ConfigError`. We expect the user to set this, either as a kwarg or in their Dask config.

The same goes for the vnet and security group.

PR #183, opened 2020-11-20
Reorder docstrings so that examples appear near the top

jacobtomlinson:
Many cluster managers have a large number of parameters and with the default autodoc ordering the examples section can get buried at the bottom.

This PR changes sphinx to preserve the ordering from the docstring and moves the examples section near the top.

I am very keen that Dask Cloudprovider has example-driven documentation. So we should work to make examples plentiful and easy to find.

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -178,6 +178,23 @@ class EC2Cluster(VMCluster):
 
     See https://docs.dask.org/en/latest/configuration.html for more info.
 
+    Examples
+    --------
+
+    Regular cluster.
+
+    >>> cluster = EC2Cluster()
+    >>> cluster.scale(5)
+
+    RAPIDS Cluster.
+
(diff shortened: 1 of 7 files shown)


jacobtomlinson:
Hrm `autodoc_member_order = "bysource"` doesn't seem to be having an effect here.

jacobtomlinson:
I'm not actually sure `autodoc_member_order` does what I want. 

Sphinx seems to always reorder the docstring sections into:

- Parameters
- Examples
- Notes
- Attributes
- Methods

I want to change this order so that examples comes first.

quasiben:
Can you list them explicitly ?

.. autoclass:: GCPCluster
   :members: Examples, Parameters, Notes....

jacobtomlinson:
I dont think so. I triend and got this error

```
WARNING: missing attribute Examples in object dask_cloudprovider.gcp.GCPCluster
WARNING: missing attribute Parameters in object dask_cloudprovider.gcp.GCPCluster
WARNING: missing attribute Attributes in object dask_cloudprovider.gcp.GCPCluster
WARNING: missing attribute Methods in object dask_cloudprovider.gcp.GCPCluster
```

I've tried a workaround which is to not use the `Examples` heading for the examples.

Issue #188, opened 2020-11-20
AzureVMCluster can.t find the provided Network Security Group


manuelreyesgomez:
Context:
Network Security Group: NGC-AML-Quick-Launch-nsg was created using:
az network nsg create -g "NGC-AML-Quick-Launch" --name "NGC-AML-Quick-Launch-nsg"
az network nsg rule create -g "NGC-AML-Quick-Launch" --nsg-name "NGC-AML-Quick-Launch-nsg" -n MyNsgRuleWithAsg --priority 500 --source-address-prefixes Internet --destination-port-ranges 8786 8787 --destination-address-prefixes '*' --access Allow --protocol Tcp --description "Allow Internet to Dask on ports 8786,8787."

The Network Security Group NGC-AML-Quick-Launch-nsg is visible in Azure Portal:
![image](https://user-images.githubusercontent.com/39012886/99833153-85437e80-2b16-11eb-990c-cc2f443cbd98.png)

Code:
from dask_cloudprovider.azure import AzureVMCluster
cluster = AzureVMCluster(resource_group="NGC-AML-Quick-Launch",vnet="NGC-AML-Quick-Launch-vnet",security_group="NGC-AML-Quick-Launch-nsg", n_workers=2,vm_size="Standard_NC12s_v3",docker_image="rapidsai/rapidsai:cuda11.0-runtime-ubuntu18.04-py3.8", worker_class="dask_cuda.CUDAWorker")

Error:
File "C:\Users\mreyesgomez\AppData\Local\Continuum\anaconda3\envs\AzureVMCluster\lib\site-packages\azure\mgmt\network\v2020_06_01\operations\_network_interfaces_operations.py", line 262, in _create_or_update_initial
    raise HttpResponseError(response=response, error_format=ARMErrorFormat)
azure.core.exceptions.HttpResponseError: (InvalidResourceReference) Resource /subscriptions/ab221ca4-f098-422d-ab2f-5073b3851e68/resourceGroups/NGC-AML-Quick-Launch/providers/Microsoft.Network/networkSecurityGroups/NGC-AML-Quick-Launch-nsg referenced by resource /subscriptions/ab221ca4-f098-422d-ab2f-5073b3851e68/resourceGroups/NGC-AML-Quick-Launch/providers/Microsoft.Network/networkInterfaces/dask-cloudprovider-nic-1699cec6 was not found. Please make sure that the referenced resource exists, and that both resources are in the same region.

Install using pip, using python 3.7

[...]

PR #191, opened 2020-11-23
Remove default location

jacobtomlinson:
There was some confusing raised in #188 due to the way we set a default location.

This PR removes the default location and adds a check to make this parameter mandatory.


nickeubank:
+1. I had this issue too.

PR #206, opened 2020-12-02
remove AzureMLCluster

lostmygithubaccount:
removes `AzureMLCluster`


mrocklin:
Woo!

jacobtomlinson:
Thanks @lostmygithubaccount. I'm tempted to leave this for a little while to give folks a chance to switch to `AzureVMCluster`. I imagine we may be doing some small bug fix releases for a while as things settle down.

lostmygithubaccount:
@jacobtomlinson sounds good - let me know when you'd like to merge and I'll update this PR as needed - thanks!

DPeterK:
I'll admit I'm against the principle of this PR, if not the implementation as such. Here's a few reasons why from a user experience perspective... 

* as @jacobtomlinson says, it's a sudden change that removes code that people might be relying upon, and who might not have time in the short term to move to the replacement. 
* it's not unlikely that a number of users of `AzureMLCluster` are people who have only been granted access to AzureML and its functionality, and not the Azure portal - that is, they can't set up any Azure resources outside of AzureML. While I recognise (and have had to work with) the challenges of the `AzureMLCluster` implementation, that's something that `AzureVMCluster` won't be able to replicate. 

Is there some way that the core functionality (I can spin up a scaling dask cluster without ever leaving AzureML) can at least be retained in some manner?

jacobtomlinson:
Thanks for joining the conversation @DPeterK. This is useful feedback and I sympathise with your requirements here.

>  it's a sudden change that removes code that people might be relying upon, and who might not have time in the short term to move to the replacement.

`AzureMLCluster` exists today under a BSD 3 license both on GitHub and via packages on PyPI and Conda Forge. Therefore it will always be possible to install and use it. Merging this PR will just remove it from future packages.

So if you want to use it you can pin your dependency version.

```
$ pip install "dask-cloudprovider<0.6"
```

Anyone is also welcome to fork the project in order to perform any bug fixes they may require.

> it's not unlikely that a number of users of `AzureMLCluster` are people who have only been granted access to AzureML and its functionality, and not the Azure portal

This is an interesting point and not something I had personally considered.

I would be very interested to learn more about how often this case comes up. Often when developing open source packages like this we only get anecdotal feedback. You are the only person so far to raise this situation.

The original author of this cluster manager is no longer with the project and therefore it is currently not being well maintained. @lostmygithubaccount has been active recently in contributing to it, but given that this PR exists it is clear this is not sustainable.

> Is there some way that the core functionality (I can spin up a scaling dask cluster without ever leaving AzureML) can at least be retained in some manner?

I see a few options here:

- Remove the code as planned. Encourage folks to move to `AzureVMCluster`. Encourage affected organisations to be less constrained with their permissions.
- Leave the currently unmaintained code in the project, potentially giving new users a poor experience due to lack of support. Accept and merge PRs from external contributors.
- Look for a new owner for that code. @DPeterK perhaps you or someone in your team may considertaking on this role?
- Write a new implementation which perhaps has a narrower scope, uses lower level AzureML primitives and is more maintainable.

This package currently has 7 different cluster managers and only 51 stars on GitHub. Therefore my assumption here is that the blast radius for removing a cluster manager at this stage will be fairly small. I am keen to get things into a more maintainable state going forwards before this ceases to be true.

lostmygithubaccount:
I'll give my perspective as well. **Disclaimer:** I am employed by Microsoft as a PM on the AI Platform/Azure Machine Learning team.

I generally agree with @jacobtomlinson - you're free to continue using `AzureMLCluster` as it's been implemented or fork and make the required improvements.

It is possible there are cases where users have access to AzureML Compute (AMLc) and not general Azure VMs, but typically the opposite it true - AMLc is less common and usually requires organizations to go through additional support requests for quota and/or approval to use.

Still, it's a valid point, and ideally I do think we should keep `AzureMLCluster` and I hope to one day add it back, but there are some key challenges:

- speed of AMLc
- development velocity of AMLv1 

**Azure ML Compute**

Azure ML is in theory agnostic to compute type, and supports:

- Azure VMs
- Azure Kubernetes (AKS)
- Azure Databricks (ADB)
- HDInsights (HDI)
- Azure Synapse Analytics (ASA) Spark Pools **soon**

Azure ML Compute is the managed compute offering, built on top of Azure Batch. It's not really recommended for bursty scaling and interactive workloads. It works will in scheduled/triggered pipelines. While we've attempted to make it work for `AzureMLCluster`, `AzureVMCluster` has better performance it should be promoted as the best user experience.

**Development velocity of AMLv1**

Azure ML, particularly `azureml-sdk`, releases every 2 weeks. For `azureml-core==1.18.0`, there were 4 patches to pypi after initially being published. There have been a number of (unintentional) breaking changes, and lots of deprecation in favor of better experiences. Some examples relevant to implementing `AzureMLCluster` include:

- `azureml.core.Image` deprecated in favor of `azureml.core.Environment`
- `azureml.core.Dataset` deprecated - see: https://docs.microsoft.com/en-us/python/api/azureml-core/azureml.core.dataset.dataset?view=azure-ml-py
- the `get_status()` method on AMLc changed return types at some point
- `azureml.train.Estimator` deprecated in favor of `azureml.core.ScriptRunConfig`

There are a number of ongoing engineering efforts which will results in further development needed of `AzureMLCluster`:

- introduction of Datasets without Datastores using AAD authentication
- introduction of `CommandJob` (which should replace `ScriptRunConfig` in `AzureMLCluster`)

I opened a PR primarily to move away from Estimators: https://github.com/dask/dask-cloudprovider/pull/144 

But with the development velocity of AMLv1 in combination with the concerns of using AMLc for this purpose, I think it's best to remove `AzureMLCluster`.

DPeterK:
Thanks @jacobtomlinson and @lostmygithubaccount for your replies to my comment! To add my own disclaimer, I work for the Met Office Informatics Lab, and we've been exploring ways of using managed ML services like Azure ML to provide Pangeo instances without having to also provision Jupyter. This is the reason for my specific interest in `AzureMLCluster` being retained - an intended audience for AzureML-backed Pangeos is organisations not dissimilar to the Met Office, where broad limitations might be placed on the Cloud resources that a given user can spin up. Thus, assuming this given user has slightly freer rein within AzureML specifically, I think there is some functionality that we stand to lose by removing the `AzureMLCluster`, even though it looks like the `AzureVMCluster` is more general and less implementationally tied to AzureML.

> fork and make the required improvements

Yes, that's always an option, and was in the back of my mind as I was writing my previous comment. With the irony fully noted, the problem remains finding someone to maintain it. I know I don't have that time and neither does anyone in the Informatics Lab, and in any case it would be nice if the code continued to be part of a repo in the dask ecosystem.

My thoughts on options for the future of this code are very similar to @jacobtomlinson's. One more option, which is probably an implementational detail on an existing suggestion, is that a mechanism is set up by which users of the `AzureVMCluster` are given specific permissions for spinning up schedulers and workers. Perhaps provider-specific examples could be provided in the docs for this repo somewhere that give examples on how this could be done as well.

I'm potentially interested in looking into a limited-scope reimplementation of `AzureMLCluster` as well - but I want to try out `AzureVMCluster` first! If it's actually fine for the use-case I'm defending here then I'll have no need to keep defending it any more 🙂

PR #207, opened 2020-12-03
Retry wait for instances

jacobtomlinson:
It seems that for some instances (particularly GPU instances) it can take a while for them to provision and even appear in `describe_instances` call.

This PR adds some logic to retry getting the instance info.

I've also implemented an exponential backoff and timeout here.


quasiben:
Thanks @jacobtomlinson. LGTM!

PR #217, opened 2020-12-11
Adding examples with EC2Cluster example

roarjn:


roarjn:
Hoping this example can be helpful to users trying to launch EC2 clusters and running ML workflow using data from S3.

jacobtomlinson:
I'm sorry that this has sat for such an extended period of time.

Issue #215, opened 2020-12-10
Allow for GCP external IPs when public_ingress is False


eric-czech:
I think this change may have been a bit extreme, or that it would be useful if it was decoupled from the `public_ingress` option: https://github.com/dask/dask-cloudprovider/pull/164/files#diff-113dfeb18ca47da584fa2c3b3d458504cc6260f0fda1055b0004f10b2b736d62R157

The use case I have is that I'm fine with VMs having external IPs, but I don't want to open up ports for anything other than ssh.  The `public_ingress=False` setting forces dask to communicate on non-ssh ports using internal ips, which is great, but the lack of an external IP means I have no way to reach hosts to debug anything.  

Personally, I think the easiest option is to change it back so the external IP is always provisioned.  Do the other VMCluster implementations ever do this conditionally?

jacobtomlinson:
Thanks for raising this @eric-czech.

This is an interesting point. Generally we make the provisioning of a public IP a boolean option. I actually think we are have made a bit of a mistake in general by having the workers connect on the public IP. I think workers should always connect privately and then the public toggle just exposes the scheduler, or not.

Out of curiosity could you use the [`gcloud` CLI](https://cloud.google.com/sdk/gcloud/reference/compute/ssh) to gain command line access to the machines without exposing port 22 at all?

jacobtomlinson:
Hrm looks like I may have already thought about this and then forgotten when implementing these new cluster managers.

dask/distributed#2963

eric-czech:
> Out of curiosity could you use the gcloud CLI to gain command line access to the machines without exposing port 22 at all?

I tried with that code change to ssh to the scheduler from another GCP VM and it didn't work.  I say "that code change" instead of without an external IP because I don't know enough to say if there are other implications of having no "compute#accessConfig" set.  Perhaps there are other options?

> Hrm looks like I may have already thought about this and then forgotten when implementing these new cluster managers.

Heh, +1 to external management and internal communication.  

A crude way I see the possibilities after reading that is:

1. Everything is public
2. Cluster hosts have management ports open to the world and no open application ports
3. Cluster hosts are not addressable outside of a VPC though certain proxy hosts are

The third option is obviously best but a bane for new users no doubt.  I suppose I'm advocating for the second option in this issue, but I can understand if it's between stools.

quasiben:
In GCP dask workers I believe always connect with internal addresses:

[...]

PR #221, opened 2020-12-14
GCP VPC

quasiben:
resolves https://github.com/dask/dask-cloudprovider/issues/215

PR allow users to define the VPC network to use when launching clusters.  For example, building a network with the following 
firewall rules restricts accessing the machines to 22, 8786, and 8787 and allows all machines in the VPC to access one another and any external address (the internet)

Screen Shot 2020-12-14 at 12 29 25 PM

cc @eric-czech

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -114,7 +114,7 @@ class GCPInstance(VMInterface):
             "networkInterfaces": [
                 {
                     "kind": "compute#networkInterface",
-                    "subnetwork": f"projects/{self.projectid}/regions/{self.general_zone}/subnetworks/{self.network or 'default'}",
+                    "subnetwork": f"projects/{self.projectid}/regions/{self.general_zone}/subnetworks/{self.network}",
                     "aliasIpRanges": [],
                 }
             ],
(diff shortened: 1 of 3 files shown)


eric-czech:
Awesome, thanks @quasiben!

dask_cloudprovider/gcp/instances.py
```diff
@@ -536,6 +542,7 @@ def __init__(
             "zone": zone or self.config.get("zone"),
             "machine_type": self.machine_type,
             "ngpus": ngpus or self.config.get("ngpus"),
+            "network": network or self.config.get("network"),
```
jacobtomlinson, line 545: I think we also need to add the default to the `cloudprovider.yaml` file.

dask_cloudprovider/gcp/instances.py
```diff
@@ -111,7 +114,7 @@ def create_gcp_config(self):
             "networkInterfaces": [
                 {
                     "kind": "compute#networkInterface",
-                    "subnetwork": f"projects/{self.projectid}/regions/{self.general_zone}/subnetworks/default",
+                    "subnetwork": f"projects/{self.projectid}/regions/{self.general_zone}/subnetworks/{self.network or 'default'}",
```
jacobtomlinson, line 117: This default value should probably live in the `cloudprovider.yaml` file.

jacobtomlinson (commented):
Thanks @quasiben for looking at this. Couple of comments.

jacobtomlinson:
Also can we add a section to the docs detaiing the rules you shared above?

jacobtomlinson (approved):
Awesome thanks!

Issue #225, opened 2020-12-20
Add monitoring scope to GCP instances


eric-czech:
API scopes for GCP instances should be configurable and/or include "monitoring-write" by default.  

https://github.com/dask/dask-cloudprovider/blob/59b142a9ef6d3c9e8838c2a0b64cf7fdc9f15919/dask_cloudprovider/gcp/instances.py#L125-L128

List of available scopes: https://cloud.google.com/sdk/gcloud/reference/alpha/compute/instances/set-scopes

This is necessary in order to use the GCP monitoring agent and without that you can't track memory utilization (along with several other important metrics).  VM images created with the monitoring agent currently fail to track due to agent errors like "The instance has neither the application default credentials file nor the correct monitoring scopes; Exiting. ... failed!".

eric-czech:
FYI I think that's the only scope necessary based on https://cloud.google.com/monitoring/agent/troubleshooting#verify-creds.  That looks to be a bit out of date, but "monitoring-write" appears to be the least permissive scope needed.

eric-czech:
I verified that this works.  Worker/scheduler instances can log metrics to the monitoring API with this change.

PR #226, opened 2020-12-20
Add monitoring.write scope for GCP #225

eric-czech:
Fixes https://github.com/dask/dask-cloudprovider/issues/225


jacobtomlinson:
Thanks so much @eric-czech

Issue #229, opened 2020-12-29
When the deploy client is within GCE use internal IP address


ravwojdyla:
This is related to #161 and #215.
The use case:
 * the deployment client is within GCE
 * we want to have public IPs (to easily download software etc, without setting up NAT gateway)
 * we don't want to configure firewall rules to allow 8787 etc traffic to external IP

To allow for this, in the client deployment code we can check if we are running in GCE and if we do, use internal IP as the address, here:
https://github.com/dask/dask-cloudprovider/blob/02b50e2efe4a98f6442e24aaedea237577cad0eb/dask_cloudprovider/gcp/instances.py#L295-L301

additionally in `wait_for_scheduler` we could add ip/port to the message to make it clear which IP the client is waited on, here:
https://github.com/dask/dask-cloudprovider/blob/02b50e2efe4a98f6442e24aaedea237577cad0eb/dask_cloudprovider/generic/vmcluster.py#L42-L45

This would streamline the deployment of GCP cluster as it removes the extra steps (firewall rules/NAT gateway). Alternatively we could also consider waiting on scheduler on both IPs at the same time, but I would lean towards simpler solution first.
FYI @eric-czech, wdyt?

eric-czech:
Sounds great!

>  we can check if we are running in GCE

Is there a standard way to do that?

> This would streamline the deployment of GCP cluster as it removes the extra steps (firewall rules/NAT gateway)

How does it do that if the instances don't have external IPs?   Without this [line](https://github.com/dask/dask-cloudprovider/pull/164/files#diff-113dfeb18ca47da584fa2c3b3d458504cc6260f0fda1055b0004f10b2b736d62R157), none of the instances I launched could reach the internet.

ravwojdyla:
@eric-czech 

>> we can check if we are running in GCE

> Is there a standard way to do that?

I mostly check if `metadata.google.internal` is available.

> How does it do that if the instances don't have external IPs? Without this line, none of the instances I launched could reach the internet.

The assumption is that we we want to have public IPs for this use case.

PR #232, opened 2020-12-30
Use internal IP whilst in GCE

ravwojdyla:
Fixes: #229

FYI @eric-czech

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -12,8 +12,7 @@ from dask_cloudprovider.generic.vmcluster import (
     VMInterface,
     SchedulerMixin,
 )
-from dask_cloudprovider.gcp.utils import build_request
-
+from dask_cloudprovider.gcp.utils import build_request, is_inside_gce
 
 from distributed.core import Status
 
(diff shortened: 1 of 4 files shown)


eric-czech:
Makes sense to me, thanks @ravwojdyla!

dask_cloudprovider/gcp/utils.py
```diff
@@ -12,3 +12,22 @@ def inner(http, *args, **kwargs):
         return googleapiclient.http.HttpRequest(new_http, *args, **kwargs)
 
     return inner
+
+
+def is_gce_env() -> bool:
```
jacobtomlinson, line 17: I'm not sure about this function name. I don't think it's immediately obvious what it does.

What about `is_inside_gce` instead?

dask_cloudprovider/gcp/instances.py
```diff
@@ -293,8 +292,16 @@ async def start_scheduler(self):
         self.internal_ip, self.external_ip = await self.create_vm()
 
         if self.config.get("public_ingress", True):
-            # scheduler is publicly available
-            self.address = f"tcp://{self.external_ip}:8786"
+            if is_gce_env():
+                # if the client is running inside GCE environment
+                # it's better to use internal IP, which doesn't
+                # require firewall setup
+                self.address = f"tcp://{self.internal_ip}:8786"
```
jacobtomlinson, line 299: We have some duplication here. Can we combine this with the outer if statement.

jacobtomlinson (commented):
Many thanks for this. I agree this change would be very valuable.

Just a couple of small comments.

dask_cloudprovider/gcp/utils.py
ravwojdyla: Sure - renamed.

dask_cloudprovider/gcp/instances.py
ravwojdyla: Good point, fixed.

dask_cloudprovider/gcp/utils.py
jacobtomlinson: Thanks!

jacobtomlinson:
Thanks for the fast review feedback turnaround!

Awesome work!

PR #234, opened 2021-01-04
Fix typo so we use subnet_id when provided, otherwise use config

sylvainv:
We cannot use the provided subnet_id provided to EC2Cluster at the moment because of a typo, this PR fixes it.


jacobtomlinson:
Many thanks!

PR #238, opened 2021-01-16
Fix tag checking in _cleanup_stale_resources and require user opt-in …

### LINKED ISSUE #94: `_cleanup_stale_resources()` deregisters unrelated ECS task definitions
We observe that Dask Cloud Provider deregisters ECS task definitions that it did not create. This screenshot shows task definitions that:

1. Dask Cloud Provider did not create
2. Do not get marked as INACTIVE (i.e. unregistered) when Dask Cloud Provider hasn't been run

image

We haven't fully characterized this, but it appears that the current version of `_cleanup_stale_resources()` is too aggressive in its de-registering of task definitions.

AndrewHannigan:
Resolves #94 by correcting usage of set membership checking, adding tag checking for Task Definition deregistration, and making default for skip_cleanup=True.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -674,7 +674,7 @@ class ECSCluster(SpecCluster):
         environment=None,
         tags=None,
         find_address_timeout=None,
-        skip_cleanup=None,
+        skip_cleanup=True,
         aws_access_key_id=None,
         aws_secret_access_key=None,
         region_name=None,
@@ -1388,7 +1388,7 @@ async def _cleanup_stale_resources():
                 )
             )["clusters"]
             for cluster in clusters:
-                if DEFAULT_TAGS.items() <= aws_to_dict(cluster["tags"]).items():
+                if set(DEFAULT_TAGS.items()) <= set(aws_to_dict(cluster["tags"]).items()):
                     if cluster["runningTasksCount"] == 0:
                         clusters_to_delete.append(cluster["clusterArn"])
                     else:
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/aws/ecs.py
```diff
@@ -674,7 +674,7 @@ def __init__(
         environment=None,
         tags=None,
         find_address_timeout=None,
-        skip_cleanup=None,
+        skip_cleanup=True,
```
martindurant, line 677: This seems like a pretty important change

dask_cloudprovider/aws/ecs.py
```diff
@@ -1434,7 +1439,7 @@ async def _cleanup_stale_resources():
                 role["Tags"] = (
                     await iam.list_role_tags(RoleName=role["RoleName"])
                 ).get("Tags")
-                if DEFAULT_TAGS.items() <= aws_to_dict(role["Tags"]).items():
+                if set(DEFAULT_TAGS.items()) <= set(aws_to_dict(role["Tags"]).items()):
```
martindurant, line 1442: So this is the real change here, right? This seems to make sense to me.

martindurant (commented):
I think @jacobtomlinson needs to check this, sorry

dask_cloudprovider/aws/ecs.py
AndrewHannigan: Yes, this fixes the core issue

dask_cloudprovider/aws/ecs.py
AndrewHannigan: I adjusted skip_cleanup to True because my thinking is that the user should have to opt-in to destructive AWS resource requests rather than have them occur by default.  But let me know your thoughts.

jacobtomlinson (changes requested):
Thanks so much for raising this. I really appreciate the fixes here.

However I don't agree with making clean up opt-in. Dask Cloudprovider should remove all resources that it creates by default. I appreciate this was a pretty serious bug, so making it opt-in feels safer, but I think we should put this back to how it was.

AndrewHannigan:
@martindurant No problem I'll revert that back

jacobtomlinson (approved):

jacobtomlinson:
Many thanks @AndrewHannigan

PR #245, opened 2021-01-27
Replace master with main

jsignell:


jacobtomlinson (changes requested):
Very much up for making the switch to `main` as the default branch here.

However, all the changes here are in references to dependencies (`pypa/gh-action-pypi-publish` and `python-versioneer/python-versioneer`) which still use `master` and therefore are not valid.

jacobtomlinson:
Given the versioneer ones are just in comments there isn't really any harm in changing them.

jsignell:
:facepalm: I didn't look close enough at the github action.

jacobtomlinson (approved):

jacobtomlinson:
Thanks @jsignell

PR #251, opened 2021-01-29
azure: add worker plugin for spot instances

### LINKED ISSUE #239: Add plugin for preemptible instances.
Hi !

Following some discussions in dask/distributed#4421. I've implemented a worker plugin that calls azure metadata service to check for node eviction in AKS and try to shut down workers as cleanly as possible. @jacobtomlinson suggested I contribute it to this repo in https://github.com/dask/distributed/issues/4421#issuecomment-760129128 somewhere in `dask_cloudprovider.azure.utils`.

Before opening an MR I would like some pointers on where to put the actual code and the associated documentation.
Thank you.

Timost:
Azure spot instances may be preempted anytime which might lead to data
losses on the dask cluster. This commit adds a worker plugin that tries
to close worker gracefully before the VM is preempted. To do so, it
polls azure metadata service to listen to premption events
notifications.

fixes #239

### DIFF
--- a/dask_cloudprovider/azure/utils.py
+++ b/dask_cloudprovider/azure/utils.py
@@ -9,6 +9,10 @@ from tornado.ioloop import IOLoop, PeriodicCallback
 
 logger = logging.getLogger(__name__)
 
+AZURE_EVENTS_METADATA_URL = (
+    "http://169.254.169.254/metadata/scheduledevents?api-version=2019-08-01"
+)
+
 
 class AzurePreemptibleWorkerPlugin(WorkerPlugin):
     """A worker plugin for azure spot instances
@@ -16,6 +20,9 @@ class AzurePreemptibleWorkerPlugin(WorkerPlugin):
     This worker plugin will poll azure's metadata service for preemption notifications.
     When a node is preempted, the plugin will attempt to shutdown gracefully all workers
     on the node.
+
+    This plugin can be used on any worker running on azure spot instances, not just the
+    ones created by ``dask-cloudprovider``.
 
     For more details on azure spot instances see:
     https://docs.microsoft.com/en-us/azure/virtual-machines/linux/scheduled-events
@@ -47,7 +54,8 @@ class AzurePreemptibleWorkerPlugin(WorkerPlugin):
     Examples
     --------
 
-    Let's say you have cluster and a client instance. For example using ``KubeCluster``
+    Let's say you have cluster and a client instance.
+    For example using :class:`dask_kubernetes.KubeCluster`
 
     >>> from dask_kubernetes import KubeCluster
     >>> from distributed import Client
@@ -55,6 +63,7 @@ class AzurePreemptibleWorkerPlugin(WorkerPlugin):
     >>> client = Client(cluster)
 
     You can add the worker plugin using the following:
+
     >>> from dask_cloudprovider.azure import AzurePreemptibleWorkerPlugin
     >>> client.register_worker_plugin(AzurePreemptibleWorkerPlugin())
     """
@@ -70,10 +79,7 @@ class AzurePreemptibleWorkerPlugin(WorkerPlugin):
         self.loop = None
         self.worker = None
         self.poll_interval_s = poll_interval_s
-        self.metadata_url = (
-            metadata_url
-            or "http://169.254.169.254/metadata/scheduledevents?api-version=2019-08-01"
-        )
+        self.metadata_url = metadata_url or AZURE_EVENTS_METADATA_URL
         self.termination_events = termination_events or ["Preempt", "Terminate"]
         self.termination_offset = datetime.timedelta(minutes=termination_offset_minutes)
 
--- a/dask_cloudprovider/azure/utils.py
+++ b/dask_cloudprovider/azure/utils.py
@@ -1,0 +1,178 @@
+import asyncio
+import datetime
+import logging
+
+import aiohttp
+from distributed.diagnostics.plugin import WorkerPlugin
+from tornado.ioloop import IOLoop, PeriodicCallback
+
+
+logger = logging.getLogger(__name__)
+
+
+class AzurePreemptibleWorkerPlugin(WorkerPlugin):
+    """A worker plugin for azure spot instances
+
+    This worker plugin will poll azure's metadata service for preemption notifications.
+    When a node is preempted, the plugin will attempt to shutdown gracefully all workers
+    on the node.
+
+    For more details on azure spot instances see:
+    https://docs.microsoft.com/en-us/azure/virtual-machines/linux/scheduled-events
+
+    Parameters
+    ----------
+    poll_interval_s: int (optional)
+        The rate at which the plugin will poll the metadata service in seconds.
+
+        Defaults to ``1``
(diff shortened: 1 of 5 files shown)


Timost:
I couldn't build the docs on my computer (import errors due to the submodule approach I think, so I'm not sure how it renders.

requirements.txt
```diff
@@ -1,3 +1,5 @@
+aiohttp>=3.7.3
 dask>=2021.01.1
 distributed>=2021.01.1
 jinja2
+tornado >= 5
```
jacobtomlinson, line 5:
```suggestion
tornado>=5
```

dask_cloudprovider/azure/utils.py
```diff
@@ -0,0 +1,178 @@
...
+        termination_events=None,
+        termination_offset_minutes=0,
+    ):
+        self.callback = None
+        self.loop = None
+        self.worker = None
+        self.poll_interval_s = poll_interval_s
+        self.metadata_url = (
+            metadata_url
+            or "http://169.254.169.254/metadata/scheduledevents?api-version=2019-08-01"
```
jacobtomlinson, line 75: This should probably be a constant defined at the top of the file.

dask_cloudprovider/azure/utils.py
```diff
@@ -0,0 +1,178 @@
...
+        Extra offset to apply to the premption date. This may be negative, to start
+        the gracefull shutdown before the ``NotBefore`` date. It can also be positive, to
+        start the shutdown after the ``NotBefore`` date, but this is at your own risk.
+
+        Defaults to ``0``
+
+    Examples
+    --------
+
+    Let's say you have cluster and a client instance. For example using ``KubeCluster``
```
jacobtomlinson, line 50: Let's make this a full reference to span the two sets of docs. Will need to add dask-kubernetes to the intersphinx mapping.

https://github.com/dask/dask-cloudprovider/blob/dffd81d3e4ab6feb5378ffa0d9952f3c7da207f7/doc/source/conf.py#L194-L198

```suggestion
    Let's say you have cluster and a client instance. For example using :class:`dask_kubernetes.KubeCluster`
```

dask_cloudprovider/azure/utils.py
```diff
@@ -0,0 +1,178 @@
...
+logger = logging.getLogger(__name__)
+
+
+class AzurePreemptibleWorkerPlugin(WorkerPlugin):
+    """A worker plugin for azure spot instances
+
+    This worker plugin will poll azure's metadata service for preemption notifications.
+    When a node is preempted, the plugin will attempt to shutdown gracefully all workers
+    on the node.
+
```
jacobtomlinson, line 19: It would be good to note here that this plugin can be used on any worker running on Azure Preemtible VMs, not just ones created by `dask-cloudprovider`.

jacobtomlinson (commented):
Thanks for this! A few comments.

Your docs probably wont build locally if you don't have all the optional dependencies installed. You can install everything with `pip install -e .[all]`. But we also have the Read The Docs PR preview builds turned on, so if you look at the CI checks you will find a link to a preview of the docs.

Timost:
Thank you for the feedback. I installed everything and now the `make html` works ( with a lot of warnings though). :heart: for the docs in the CI.
I'll make the changes and update the PR !

TomAugspurger:
This is neat to see. If we're able to, it'd be great to determine what parts are specific to Azure VMs, and what applies to other Azure services like AKS. If this works with AKS as well, then I think it'd be appropriate to include the plugin in `distributed` instead of `dask-cloudprovider`.

Timost:
I'm actually using this with AKS (with vmss backed spot node pools) not Azure VMs, in fact I haven't tested it with Azure VMs.

If the consensus is to put this in `distributed` I don't mind moving the code over there. I'm don't have a strong opinion on this, except that `dask-cloudprovider` is quite new and every `dask` user might not be aware it exists.

jacobtomlinson:
> If this works with AKS as well, then I think it'd be appropriate to include the plugin in distributed instead of dask-cloudprovider so that dask-gateway and dask-kubernetes can benefit from it as well.

It should work on everything preemptible in Azure.

We did discuss this over in distributed, but felt that cloudprovider is the appropriate place for cloud specific things. There is nothing stopping you from making use of this with other libraries. 

Perhaps we should consider adding references to it in other docs.

TomAugspurger:
Mmm, was that discussed in https://github.com/dask/distributed/issues/3300?

I'd find it a bit strange to install dask-cloudprovider if you're *only* using it for this plugin (if you're using dask-kubernetes, say).

While it *is* cloud-specific, I think that since multiple sub-projects would get value out of it, `distributed` would make a good home, since each of the sub-projects depends on distributed.

Timost:
It was discussed in https://github.com/dask/distributed/issues/4421. Sorry it's kind of hard to follow. The original issue ended up raising several other issues on different projects.

I understand both points of view, I think `dask-cloudprovider` has several submodules that can be installed separately. So, another solution could be putting the plugin in a separate module that can be installed independently of the rest of `dask-cloudprovider`. What do you think ? Maybe that's too much modules...

jacobtomlinson:
@TomAugspurger I see where you are coming from.

My thinking here was that these plugins may have cloud dependencies (although I don't think this one does).

So installing `dask-cloudprovider` (or specifically `pip install dask-cloudprovier[azure]`) to get cloud specific libraries and dependencies makes sense.

The cloud vendor dependencies can be very heavy and we install them here only as extras. I would like to avoid that code going into distributed.

TomAugspurger:
Yeah, the dependency point is strong. I agree that when there are cloud-specific dependencies, it's best to have them in a package that depends on them rather than relying on runtime checks. Since this plugin doesn't do anything with azure.*, my thought is it would be better suited for distributed (aiohttp would need to be replaced with tornado.httpclient).

But I don't know what other cloud-provider specific plugins are available in dask-cloudprovider, and it would be a bit strange to have some in distributed and some in dask-cloudprovider, depending on their implementation details.

Timost:
> (aiohttp would need to be replaced with tornado.httpclient).

Thank you for pointing this out, I'll keep it in mind.

> But I don't know what other cloud-provider specific plugins are available in dask-cloudprovider, and it would be a bit strange to have some in distributed and some in dask-cloudprovider, depending on their implementation details.

AFAIK there are no other cloud-provider specific plugins in `dask-cloudprovider` for now.

I'll soon get the opportunity to run the same pipeline on google's GKE, so I thought I might mention it here as it would have triggered writing an equivalent plugin for gcp. But it seems google's preemption notification is done entirely differently (https://cloud.google.com/compute/docs/instances/preemptible#preemption-process).

TomAugspurger:
Don't let me comments about where this should live derail the implementation here. If we decide that distributed is a better home in the future then we can make the migration then.

Timost:
PTAL

jacobtomlinson:
Ok this looks great.

I'm going to get this merged in here now and also open an issue on Distributed to discuss upstreaming it. I'll raise it at the Dask maintainers meeting too.

Timost:
:+1: Thank you !

Issue #255, opened 2021-02-09
Add support for GCP preemptible instances


tomwhite:
It would be useful to be able to try some workloads on a cluster with preemptible workers.

Ideally the worker would try to shut down gracefully, but I've successfully some workflows without this (see https://github.com/pystatgen/sgkit/issues/453). Also, it's not clear if the 30s termination warning is enough time to save the worker memory state in practice, although it would still be worth the worker shutting down on a best effort basis.

tomwhite:
Fixed in #256

PR #256, opened 2021-02-09
Add option for preemptible instances on GCP

tomwhite:
Here's a simple fix for #255.

As it stands, if preemptible instances are enabled, all instances in the cluster - including the scheduler - are preemptible. It would be better if the scheduler instance was always a standard (non-preemptible) instance. Suggestions for how to do that would be very welcome.

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -61,7 +61,7 @@ class GCPInstance(VMInterface):
         bootstrap=None,
         gpu_instance=None,
         auto_shutdown=None,
-        preemptible=None,
+        preemptible=False,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -571,7 +571,9 @@ class GCPCluster(VMCluster):
             "gpu_instance": self.gpu_instance,
             "bootstrap": self.bootstrap,
             "auto_shutdown": self.auto_shutdown,
-            "preemptible": preemptible if preemptible is not None else self.config.get("preemptible"),
+            "preemptible": preemptible
+            if preemptible is not None
+            else self.config.get("preemptible"),
         }
         self.scheduler_options = {**self.options}
         self.worker_options = {**self.options}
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/gcp/instances.py
```diff
@@ -146,9 +148,9 @@ def create_gcp_config(self):
             },
             "labels": {"container-vm": "dask-cloudprovider"},
             "scheduling": {
-                "preemptible": "false",
+                "preemptible": ("true" if self.preemptible else "false"),
```
quasiben, line 151: What do you think about making the default  `preemptible=False` rather than None ?  I think keeping the massive dictionary cleaner is a good goal

quasiben (commented):
Thank @tomwhite !  This looks great.  Left one small comment.  I think it would also be good if there was a test for preemptible instances.  Lastly, will GCP reject the submission if `preemptible` is True and `automaticRestart` is True?

dask_cloudprovider/gcp/instances.py
tomwhite: Fixed

tomwhite:
Thanks for the review @quasiben.

I haven't added a test yet - what would you recommend?

Also, any thoughts on how to make the scheduler non-preemptible?

> will GCP reject the submission if `preemptible` is True and `automaticRestart` is True?

Yes, it did when I tried.

quasiben:
> I haven't added a test yet - what would you recommend?

Naively, a test where `preemptible=True` -- thinking about this now, it's probably not worth it given how long it already takes to run the test suite. Last question according to [google docs](https://cloud.google.com/compute/docs/instances/preemptible#preemptible-with-gpu) I think preemptible instances should just work, is that correct ?

tomwhite:
I've updated with a change so that scheduler instances are never preemptible. This is important, since losing the scheduler means the whole cluster is unusable.

> Last question according to [google docs](https://cloud.google.com/compute/docs/instances/preemptible#preemptible-with-gpu) I think preemptible instances should just work, is that correct ?

Are you asking about GPUs on preemptible instances? I haven't tried them, but that doc suggests they would work the same.

quasiben:
Sorry, yes. I was asking about GPU instances.  The last change where schedulers are not preemptible is a very good idea.  Merging now.  

Thanks @tomwhite !

PR #274, opened 2021-03-24
[GCP] Add disk type config option

### LINKED ISSUE #272: ssd option for GCPCluster
It would be nice to be able to specify ssds for the nodes when you instantiate a GCPCluster. For example, when you create a cluster in gcloud, you can pass parameters like

    --master-boot-disk-type pd-ssd
    --worker-boot-disk-type pd-ssd

jacobtomlinson:
Exposed disk type as an option for `GCPCluster`.

Closes #272


quasiben (approved):
Thanks @jacobtomlinson !

PR #285, opened 2021-05-10
Update documentation for required FargateCluster IAM Roles

JamesJeffryes:
Thanks for this awesome library! I was spinning up a Fargate Cluster with a custom policy based on the roles in the documentation and ran into a few missing IAM roles. I've added the ones I had to add to the docstring.

Note that I only tested this on a cluster with no specified resources but I'm reasonably confident that they would be needed for a one where resources are specified too.


jacobtomlinson (approved):
This is great many thanks!

Issue #293, opened 2021-06-10
[BUG] Cannot create VM using Azure Marketplace VM Image with plan via AzureVMCluster


akaanirban:
**What happened**:
Trying to create an AzureVMCluster fails if we use an Azure Marketplace VM (NVIDIA GPU-Optimized Image for AI & HPC) with a plan. This needs some plan information to be passed along while creating the VMs in a `plan` field. This field is not passed in the `vm_parameters`. Hence it is failing. 

<details>
<summary>The actual error. </summary>
<p>

```bash
---------------------------------------------------------------------------
OperationFailed                           Traceback (most recent call last)
/opt/conda/envs/rapids/lib/python3.8/site-packages/azure/core/polling/base_polling.py in run(self)
    481         try:
--> 482             self._poll()
    483 

/opt/conda/envs/rapids/lib/python3.8/site-packages/azure/core/polling/base_polling.py in _poll(self)
    520         if _failed(self.status()):
--> 521             raise OperationFailed("Operation failed or canceled")
    522 

OperationFailed: Operation failed or canceled

During handling of the above exception, another exception occurred:

HttpResponseError                         Traceback (most recent call last)
<timed exec> in <module>

/rapids/notebooks/host/dask-cloudprovider/dask_cloudprovider/azure/azurevm.py in __init__(self, location, resource_group, vnet, security_group, public_ingress, vm_size, scheduler_vm_size, vm_image, disk_size, bootstrap, auto_shutdown, docker_image, debug, is_marketplace_vm, **kwargs)
    545         }
    546         self.worker_options = {"vm_size": self.vm_size, **self.options}
--> 547         super().__init__(debug=debug, **kwargs)

/rapids/notebooks/host/dask-cloudprovider/dask_cloudprovider/generic/vmcluster.py in __init__(self, n_workers, worker_class, worker_options, scheduler_options, docker_image, docker_args, env_vars, security, protocol, debug, **kwargs)
    287         self.uuid = str(uuid.uuid4())[:8]
    288 
--> 289         super().__init__(**kwargs, security=self.security)
    290 
    291     async def call_async(self, f, *args, **kwargs):

/opt/conda/envs/rapids/lib/python3.8/site-packages/distributed/deploy/spec.py in __init__(self, workers, scheduler, worker, asynchronous, loop, security, silence_logs, name, shutdown_on_close)
    280         if not self.asynchronous:
    281             self._loop_runner.start()
--> 282             self.sync(self._start)
[...]

PR #294, opened 2021-06-10
Added support for using Marketplace VM images with plans in AzureVMCluster

akaanirban:
This PR :

- Adds support for using Marketplace VM images that has plans while creating AzureVM Cluster. 
- Closes [Issue#293](https://github.com/dask/dask-cloudprovider/issues/293) 

**More Details on Azure Marketplace VM issue**

Creating a virtual machine from Marketplace image or a custom image sourced from a Marketplace image requires Plan information in the request when using the python sdk or cli. This is applicable to the images which has plans. For example the pre-build *NVIDIA GPU-Optimized Image for AI & HPC* image made available by NVIDIA.

As an example, the input fields for creating a VM using *NVIDIA GPU-Optimized Image for AI & HPC* 21.02.2 can be found using the following: 
```bash
 >> az vm image show --publisher nvidia --offer ngc_azure_17_11 --sku ngc-base-version-21-02-2 --version 21.02.2
{
  "automaticOsUpgradeProperties": {
    "automaticOsUpgradeSupported": false
  },
  "dataDiskImages": [],
  "disallowed": {
    "vmDiskType": "Unmanaged"
  },
  "extendedLocation": null,
  "features": [
    {
      "name": "IsAcceleratedNetworkSupported",
      "value": "True"
    }
  ],
  "hyperVGeneration": "V1",
  "id": "/Subscriptions/<truncated subscription id>/Providers/Microsoft.Compute/Locations/westus/Publishers/nvidia/ArtifactTypes/VMImage/Offers/ngc_azure_17_11/Skus/ngc-base-version-21-02-2/Versions/21.02.2",
  "location": "westus",
  "name": "21.02.2",
  "osDiskImage": {
    "operatingSystem": "Linux",
    "sizeInBytes": 34359738880,
    "sizeInGb": 32
  },
  "plan": {
    "name": "ngc-base-version-21-02-2",
    "product": "ngc_azure_17_11",
    "publisher": "nvidia"
  },
  "tags": null
}
```
The `plan` field is an additional field that needs to be passed in this case. The contains of this field is essentially similar to what we have in `vm_image` field, just the names of the keys are different. 

**Soluton:**
I introduced a flag `is_marketplace_vm_with_plan`, which is by default False. You should set it True if you want to use a marketplace VM image with a plan. If set to True, this will extract the `name` , `product` and `publisher` information from `vm_image` and sets to a upper level key `plan`. 

**Caveat**
Inferring if the user passed in image is a marketplace image or if the market place image has a plan or not is not automatic. If the user accidentally sets the `is_marketplace_vm_with_plan` to True and don't use a Marketplace VM or an image with a Plan, then it would throw an error.

### DIFF
--- a/dask_cloudprovider/azure/azurevm.py
+++ b/dask_cloudprovider/azure/azurevm.py
@@ -43,7 +43,7 @@ class AzureVM(VMInterface):
         env_vars: dict = {},
         bootstrap: bool = None,
         auto_shutdown: bool = None,
-        is_marketplace_vm_with_plan: bool = False,
+        marketplace_plan : dict = {},
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -65,7 +65,7 @@ class AzureVM(VMInterface):
         self.disk_size = disk_size
         self.auto_shutdown = auto_shutdown
         self.env_vars = env_vars
-        self.is_marketplace_vm_with_plan = is_marketplace_vm_with_plan
+        self.marketplace_plan = marketplace_plan
 
     async def create_vm(self):
         [subnet_info, *_] = await self.cluster.call_async(
@@ -158,14 +158,15 @@ class AzureVM(VMInterface):
             "tags": self.cluster.get_tags(),
         }
 
-        if self.is_marketplace_vm_with_plan:
+        if self.marketplace_plan:
+            # Ref: https://docs.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update#create-a-vm-with-a-marketplace-image-plan.
+            # Creating a marketplace VM with a plan will override default vm_image values.
+            vm_parameters["plan"] = self.marketplace_plan
+            vm_parameters["storage_profile"]["image_reference"]["sku"] = self.marketplace_plan["name"]
+            vm_parameters["storage_profile"]["image_reference"]["publisher"] = self.marketplace_plan["publisher"]
+            vm_parameters["storage_profile"]["image_reference"]["offer"] = self.marketplace_plan["product"]
+            vm_parameters["storage_profile"]["image_reference"]["version"] = "latest"
             self.cluster._log("Using Marketplace VM image with a Plan")
-            plan = {
-                "name": self.vm_image["sku"],
-                "publisher": self.vm_image["publisher"],
-                "product": self.vm_image["offer"]
-            }
-            vm_parameters["plan"] = plan
 
         self.cluster._log("Creating VM")
         if self.cluster.debug:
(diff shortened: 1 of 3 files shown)


jacobtomlinson:
Thanks @akaanirban.

We should also expose this as a Dask config option. See other uses of `dask.config.get` for examples.

https://github.com/dask/dask-cloudprovider/blob/6a629089b583469f70879943d7f127b77c315736/dask_cloudprovider/cloudprovider.yaml#L55-L72

akaanirban:
Thanks @jacobtomlinson  for the suggestion! I think something like this may work in the `yaml` file
```yaml
azure:
    location: null # The Azure location to launch your cluster
    resource_group: null # The Azure resource group for the cluster
    azurevm:
      vnet: null # Azure Virtual Network to launch VMs in
      security_group: null # Network security group to allow 8786 and 8787
      public_ingress: true # Assign a public IP address to the scheduler
      vm_size: "Standard_DS1_v2" # Azure VM size to use for scheduler and workers
      disk_size: 50  # Specifies the size of the VM host OS disk in gigabytes. This value cannot be larger than `1023`.
      scheduler_vm_size: null # Set a different VM size for the scheduler. Will use vm_size if not set
      vm_image: # OS image to use for the virtual machines
        publisher: "Canonical"
        offer: "UbuntuServer"
        sku: "18.04-LTS"
        version: "latest"
      bootstrap: true # It is assumed that the VHD does not have Docker and needs bootstrapping. Set this to false if using a custom VHD with Docker already installed.
      auto_shutdown: true # Shutdown instances automatically if the scheduler or worker services time out.
      is_marketplace_vm_with_plan: false # This is to be passed true only when a marketplace vm with a plan is used
      plan:  #only used when is_marketplace_vm_with_plan is set to True, defaults to a basic free NVIDIA GPU optimized VM plan.
        name: "ngc-base-version-21-02-2"
        publisher: "nvidia"
        product: "ngc_azure_17_11"
```

I think the logic should now be:
```bash
if `is_marketplace_vm_with_plan` == True:
    if user has provided a `plan` information via dask config:
        use that
    else if user has provided a vm_image information:
        try to convert it to plan information and use that (what I did currently)
    else:
        use the default plan information in the dask config.  
else:
    use normal path as before using `vm_image` information
```
1. Can you verify if that should be the correct flow as per dask-cloudprovider ?
2. What if the user accidentally sets the `is_marketplace_vm_with_plan` to true and passes in a `vm_image` which does not need a plan? I thought that getting an error in such a scenario is correct, since the error would be from Azure's side.

jacobtomlinson:
I would be tempted to change the config to be

```yaml
      marketplace_plan:  null 
        # name: "ngc-base-version-21-02-2"
        # publisher: "nvidia"
        # product: "ngc_azure_17_11"
```

And have a `marketplace_plan` kwarg instead of `is_marketplace_vm_with_plan` which defaults to `None` and takes a dict of values.

In the `__init__` you need to do 

```python
self.marketplace_plan = marketplace_plan or self.config.get("marketplace_plan")
```

Then in the create VM method you can do

```python
        if self.marketplace_plan:
            # Check that self.marketplace_plan contains the right options
            vm_parameters["plan"] = self.marketplace_plan
```

You would need

akaanirban:
@jacobtomlinson  Actually that makes much more sense. Users won't need to set two things then! If the user sets the plan information via dask config or via `kwargs` then it would receive precedence. WIll make the changes and update :D

jacobtomlinson:
This looks great many thanks! Have you been able to successfully run the full Azure tests locally? We don't currently have CI set up here for anything which creates cloud resources.

jacobtomlinson:
Chatted with @akaanirban offline. Tests are ok.

jacobtomlinson (approved):

jacobtomlinson:
Thanks @akaanirban

PR #298, opened 2021-07-01
[Azure] Improve cluster config interface

### LINKED ISSUE #297: AzureVMCluster doesn't correctly load config
**What happened**:
When I defined a default set of arguments for `AzureVMCluster` in `~/.config/dask/cloudprovider.yaml` and tried to create a cluster instance with no configuration arguments, I received an error.

**What you expected to happen**:
I would have expected to get an `AzureVMCluster` instance matching the config specified in the yaml file.

**Minimal Complete Verifiable Example**:

I set the following config in `~/.config/dask/cloudprovider.yaml`:

```yaml
cloudprovider:
  azure:
    location: "westeurope" # The Azure location to launch your cluster
    resource_group: "HelloML" # The Azure resource group for the cluster
    azurevm:
      vnet: "dask-cp-vnet" # Azure Virtual Network to launch VMs in
      security_group: "dask-cp-nsg" # Network security group to allow 8786 and 8787
      vm_size: "Standard_D2_v3" # Azure VM size to use for scheduler and workers
```
This should be sufficient & valid config to create an `AzureVMCluster` instance, as I have constructed other clusters using the same config but passed as kwargs to the constructor.

However, when I try to create a cluster without specifying this config again, for example:
```python
avmc = AzureVMCluster(n_workers=1)
```

I get the following error:
```python-traceback
---------------------------------------------------------------------------
ConfigError                               Traceback (most recent call last)
<ipython-input-2-558400e25ae3> in <module>
----> 1 avmc = AzureVMCluster(n_workers=1)

/anaconda/envs/tiledb-dev/lib/python3.8/site-packages/dask_cloudprovider/azure/azurevm.py in __init__(self, location, resource_group, vnet, security_group, public_ingress, vm_size, scheduler_vm_size, vm_image, disk_size, bootstrap, auto_shutdown, docker_image, debug, marketplace_plan, **kwargs)
    477         )
    478         if self.resource_group is None:
--> 479             raise ConfigError("You must configure a resource_group")
    480         self.public_ingress = (
    481             public_ingress

ConfigError: You must configure a resource_group
```

**Anything else we need to know?**:
Looking at `AzureVMCluster.__init__` there seem to be a couple of different approaches taken to get config items:
* `dask.config.get("cloudprovider.azure.location")` ([L469](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/azure/azurevm.py#L469))
* `self.config.get("resource_group")` ([L476](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/azure/azurevm.py#L476) etc.)

Changing L476 to follow the pattern of L469 (that is, so that it reads `dask.config.get("cloudprovider.azure.resource_group")`) fixed the problem for me, which suprised me a little bit as I might have expected to need to change all the config getters to follow the pattern of L469. I assume that the difference is due to the levels within the yaml file that provide each config item, with items at the `cloudprovider.azure` level needing to be retrieved from `dask.config`, but items within the sub-level `cloudprovider.azure.azurevm` are automatically being loaded as part of the `AzureVMCluster` instantiation?

I'm happy to raise this change as a PR, but I wanted to check first that my assumption above was correct! For example, assumedly another route to fixing this would be to change where the `resource_group` item is defined within the yaml file, but I don't know if that would have other unexpected impacts.

**Environment**:

- Dask version: 2021.06.2
- Python version: 3.8
- Operating System: AzureML Linux (Ubuntu 18.10 I think)
- Install method (conda, pip, source): conda

DPeterK:
Make the `config` interfaces in `AzureVMCluster` consistently reference `AzureVMCluster.config` rather than sometimes referring out to `dask.config` itself. 

This also:
* enables nested config access via `.` notation, as per `dask.config.get`
* fixes a bug with accessing the `resource_group` key from `dask.config`.

Fixes #297 - see also the discussion there on the best approach for this fix.

### DIFF
--- a/dask_cloudprovider/azure/azurevm.py
+++ b/dask_cloudprovider/azure/azurevm.py
@@ -467,10 +467,14 @@ class AzureVMCluster(VMCluster):
         self.location = self.config.get("location", override_with=location)
         if self.location is None:
             raise ConfigError("You must configure a location")
-        self.resource_group = self.config.get("resource_group", override_with=resource_group)
+        self.resource_group = self.config.get(
+            "resource_group", override_with=resource_group
+        )
         if self.resource_group is None:
             raise ConfigError("You must configure a resource_group")
-        self.public_ingress = self.config.get("azurevm.public_ingress", override_with=public_ingress)
+        self.public_ingress = self.config.get(
+            "azurevm.public_ingress", override_with=public_ingress
+        )
         self.credentials, self.subscription_id = get_azure_cli_credentials()
         self.compute_client = ComputeManagementClient(
             self.credentials, self.subscription_id
@@ -481,7 +485,9 @@ class AzureVMCluster(VMCluster):
         self.vnet = self.config.get("azurevm.vnet", override_with=vnet)
         if self.vnet is None:
             raise ConfigError("You must configure a vnet")
-        self.security_group = self.config.get("azurevm.security_group", override_with=security_group)
+        self.security_group = self.config.get(
+            "azurevm.security_group", override_with=security_group
+        )
         if self.security_group is None:
             raise ConfigError(
                 "You must configure a security group which allows traffic on 8786 and 8787"
@@ -493,8 +499,7 @@ class AzureVMCluster(VMCluster):
                 "VM OS disk canot be larger than 1023. Please change the ``disk_size`` config option."
             )
         self.scheduler_vm_size = self.config.get(
-            "azurevm.scheduler_vm_size",
-            override_with=scheduler_vm_size
+            "azurevm.scheduler_vm_size", override_with=scheduler_vm_size
         )
         if self.scheduler_vm_size is None:
             self.scheduler_vm_size = self.vm_size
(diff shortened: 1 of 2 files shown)


dask_cloudprovider/azure/azurevm.py
```diff
@@ -241,6 +241,19 @@ class AzureVMWorker(WorkerMixin, AzureVM):
...
+    Enables '.' notation for nested access, as per `dask.config.get`.
+
+    """
+    def __new__(cls, d):
+        return super().__new__(cls, d)
+        
+    def get(self, key, default=None, override_with=None):
+        return dask.config.get(key, default=default, config=self, override_with=override_with)
+
+
```
jacobtomlinson, line 256: This would be useful in a bunch of places, would you mind moving it somewhere like `dask_cloudprovider.utils` or maybe even `dask_cloudprovider.config`?

Also happy for it to be a public class.

dask_cloudprovider/azure/azurevm.py
```diff
@@ -489,21 +502,21 @@ def __init__(
...
             if security_group is not None
-            else self.config.get("security_group")
+            else self.config.get("azurevm.security_group")
         )
         if self.security_group is None:
             raise ConfigError(
                 "You must configure a security group which allows traffic on 8786 and 8787"
             )
-        self.vm_size = vm_size if vm_size is not None else self.config.get("vm_size")
+        self.vm_size = vm_size if vm_size is not None else self.config.get("azurevm.vm_size")
```
jacobtomlinson, line 517: While you're here we could simplify each of these using the `override_with` kwarg that you've helpfully exposed in the custom config object.

```suggestion
        self.vm_size = self.config.get("azurevm.vm_size", override_with=vm_size)
```

jacobtomlinson (commented):
Thanks so much @DPeterK. A couple of comments.

dask_cloudprovider/azure/azurevm.py
DPeterK: Oh yeah 🤦

dask_cloudprovider/azure/azurevm.py
DPeterK: Sure thing. Do you also want more detailed documentation for the class?

jacobtomlinson: Perhaps just a small `Example` block would be helpful for future implementers (probably me).

jacobtomlinson:
This is pleasing 

image

jacobtomlinson (approved):
This is great thanks @DPeterK

PR #305, opened 2021-08-13
Check for pending tasks during _cleanup_stale_resources for ECS

bbertran:
The code would try and delete clusters with pending tasks resulting in failure.  Added check for 0 pending tasks as well as running tasks.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -1404,7 +1404,7 @@ async def _cleanup_stale_resources():
                 if set(DEFAULT_TAGS.items()) <= set(
                     aws_to_dict(cluster["tags"]).items()
                 ):
-                    if cluster["runningTasksCount"] == 0:
+                    if cluster["runningTasksCount"] == 0 and cluster["pendingTasksCount"] == 0:
                         clusters_to_delete.append(cluster["clusterArn"])
                     else:
                         active_clusters.append(cluster["clusterName"])


lmarsden:
We have seen the issue described in our systems. We currently set `skip_cleanup` to `True` when instantiating ECS/Fargate clusters as a workaround.

jacobtomlinson (commented):
Many thanks for raising this.

So am I right in thinking that the cluster is being destroyed while some tasks are still being created? If so this change will just skip over the deletion right?

This probably isn't ideal if you create a cluster and then quickly delete it again as it won't actually be deleted. Would it be better to wait or maybe fail instead?

bbertran:
The change makes it such that it will not try to delete a cluster that has tasks in pending state.  In our use case, we are launching prefect flows using dask-cloudprovider.  If we launch multiple flows at the same time with skip_cleanup = False we will hit the exception shown below as it attempts to delete clusters which have not yet started running.  

I'm fairly new to dask so your suggestions might be better suited for more use cases.  From our point of view it didn't make sense to delete a cluster with a task still pending (it was pulling a container image).

botocore.errorfactory.ClusterContainsTasksException: An error occurred (ClusterContainsTasksException) when calling the DeleteCluster operation: The Cluster cannot be deleted while Tasks are active.

jacobtomlinson:
> The change makes it such that it will not try to delete a cluster that has tasks in pending state.

Sure, which means if you create a cluster and immediately delete it then it won't actually be deleted. I can see this resulting in users spending money on clusters they thought they had closed.

It would be better to cancel the pending tasks and then closing out.

lmarsden:
As the code change is in `_cleanup_stale_resources()` I don't think it would affect explicit deletion of newly created clusters shortly after creation. It targets resources considered 'stale' at cluster start-up. Without this change, if >1 clusters are instantiated in quick succession they are oftentimes considered stale, because they have no running tasks, and torn down in error along with genuinely stale clusters that also have no running tasks (and no pending tasks either).

bbertran:
I'm not quite following your point on immediate deletion.  From my understanding with the current implementation, trying to delete a pending cluster would result in the exception shown in my last comment and it would not be deleted.  This change would just prevent the exception / failure.  If it were changed to cancel/delete the pending tasks that would not be ideal for us as we are trying to launch multiple clusters at once and this would result in only the last cluster running.  Currently we are just skipping the cleanup, but it would be nice to have it managed automatically.

On a side note, we are also hitting the 10 second timeout issue when using Fargate so our ECS clusters are not being deleted after completion (https://github.com/dask/dask-cloudprovider/issues/291)

jacobtomlinson:
Right, I'm with you! Apologies for the confusion. In that case yes I agree this makes sense.

PR #307, opened 2021-08-24
Fix an import issue introduced by https://github.com/aio-libs/aioboto…

### LINKED ISSUE #306: a new PR in aoibotocore introduced a bug in AWS related code
Running dask_cloudprovider with the latest version of aiobotocore (installed via pip) starting a fargate ecs dask cluster result in an exception:

```
  File "/usr/local/lib/python3.6/site-packages/dask_cloudprovider/aws/ecs.py", line 1361, in __init__
    super().__init__(fargate_scheduler=True, fargate_workers=True, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/dask_cloudprovider/aws/ecs.py", line 726, in __init__
    self.session = aiobotocore.get_session()
AttributeError: module 'aiobotocore' has no attribute 'get_session'
```

gmorin:
A pull request in aiobotocore (https://github.com/aio-libs/aiobotocore/pull/877) removed the get_session alias from the main package.

This will close #306

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -17,7 +17,7 @@ from dask_cloudprovider.aws.helper import (
 from dask_cloudprovider.utils.timeout import Timeout
 
 try:
-    import aiobotocore
+    from aiobotocore.session import get_session
     import botocore.exceptions
 except ImportError as e:
     msg = (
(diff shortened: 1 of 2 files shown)


thehesiod:
FYI 1.4.1 has put this change behind an env var to not break backwards compatibility.  Also the other change with that are the http exceptions thrown. With the var set to 0, we no longer throw raw aiohttp exceptions, and instead throw the wrapped botocore exceptions for parity

jacobtomlinson (commented):
Thanks for the fix here @gmorin.

@thehesiod thanks for the info. Am I right in thinking this fix will be backward compatible anyway?

thehesiod:
@jacobtomlinson correct

jacobtomlinson (approved):
Thanks for confirming @thehesiod.

This is in.

Issue #261, opened 2021-03-05
Option to not specify launchType in ECS task


kinghuang:
When using ECSCluster, the generated ECS task includes `"launchType": "FARGATE" if self.fargate else "EC2"` ([ec2.py#L247](https://github.com/dask/dask-cloudprovider/blob/adca69c94348181daa4f56cac8ab9597081483bd/dask_cloudprovider/aws/ecs.py#L247)). This makes it impossible to use the default capacity provider for an ECS cluster, by not specifying `capacityProviderStrategy` or `launchType`.

From [RunTask](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_RunTask.html) API docs:

> If no `capacityProviderStrategy` or `launchType` is specified, the defaultCapacityProviderStrategy for the cluster is used.

The current implementation also makes it impossible to specify a `capacityProviderStrategy` via `task_kwargs`, because `capacityProviderStrategy` and `launchType` cannot be specified together.

> If a `capacityProviderStrategy` is specified, the `launchType` parameter must be omitted.

I would like a way to *not* specify `launchType` with ECSCluster, so that I can either use the cluster's default capacity provider or specify a `capacityProviderStrategy`.

jacobtomlinson:
Thanks for raising this @kinghuang. That sounds like a reasonable request.

Do you have any interest in raising a Pull Request to resolve this?

kinghuang:
Interest, yes. But, lacking in available time. :)

I'll circle back to this later in the month, if no one else picks it up!

kinghuang:
I'll pick this up on Wednesday.

PR #312, opened 2021-10-02
[Dask] Add support for AWS Fargate Spot clusters

dirtyValera:
Runs Workers on `FARGATE_SPOT` and Scheduler on `FARGATE` capacity providers if `FargateCluster(fargate_spot=True)`

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -74,7 +74,7 @@ class Task:
         log group when launching this task.
 
     fargate: bool
-        Whether or not to launch with the Fargate launch type.
+        Whether or not to launch on Fargate.
 
     environment: dict
         Environment variables to set when launching the task.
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
Thanks for raising this. 

Could you share a little on the decision to add `fargate_capacity_provider` and `fargate_spot` kwargs and what the difference is?

Also it looks like linting is failing.

dirtyValera:
> Thanks for raising this.
> 
> Could you share a little on the decision to add `fargate_capacity_provider` and `fargate_spot` kwargs and what the difference is?
> 
> Also it looks like linting is failing.

`fargate_spot` indicates a cluster level configuration: scheduler on `FARGATE` and workers on `FARGATE_SPOT` capacity providers. Since it is a cluster logic, I didn't pass the flag down to workers (`FARGATE` vs `FARGATE_SPOT` decision is made outside `Task` class).

`fargate_capacity_provider` is a capacity provider specific for `fargate_spot=True` configuration since it is set only when tasks run on Fargate Spot. I didn't generalize to setting arbitrary capacity provider to make sure it's not a breaking change (e.g. https://github.com/dask/dask-cloudprovider/issues/261, some people already pass down capacity providers in kwargs, I wanted to keep it as it is, defaulting to `fargate_spot=False`)

jacobtomlinson:
Thanks @dirtyValera could you update the docstrings to make it clear how the two options interact?

dirtyValera:
> Thanks @dirtyValera could you update the docstrings to make it clear how the two options interact?

Updated

jacobtomlinson (approved):
Much appreciated!

tracek:
Hey @dirtyValera - I was wondering if you know if I might be doing something incorrectly when trying to use this nice feature you have implemented: https://github.com/dask/dask-cloudprovider/issues/328 ? Is this something you can repro or is it just me? If it's general, I could try to fix it & PR. Maybe you already have a good idea what could be the issue?

Issue #320, opened 2021-11-30
Allow choosing number of threads in ECS workers


drorspei:
The default number of threads per ECS worker task is the number vCPUs.

I'm using ECSCluster to start workers that run tasks which already use threading, so I would like the class to take a `worker_nthreads` argument, similarly to LocalCluster, so that I could set it to something lower than the current default.

What do you think?

PR #321, opened 2021-11-30
New argument `worker_nthreads` in ECSCluster

drorspei:
The default number of threads in worker tasks is the number of vCPUs.

This pull request adds an argument `worker_nthreads` so that users can override the default. This is issue #320 .


jacobtomlinson (approved):
This looks reasonable thanks!

Issue #322, opened 2021-11-30
Allow passing an existing scheduler to ECSCluster


drorspei:
The currently supported workflow is to start an ECSCluster before doing any dask work, and only once it's started start doing work.

It would be very convenient if we could pass an existing scheduler address, not start a scheduler task, and the workers would connect to the passed address. 

How about it?

PR #323, opened 2021-11-30
Add arguments so ECSCluster supports an existing scheduler

drorspei:
The currently supported workflow is to start an ECSCluster before doing any dask work, and only once it's started start doing work.

Recently I have moved to the following workflow: start a dask LocalCluster, start working, and if I decide I need more computing power then I spin up an ECSCluster that connects to the existing Scheduler. I do this from a separate python process, maybe even a different machine, and I watch the number of workers grow and the tasks finish in the dashboard. This also decouples the computing resources from the code actually using dask.

Making the cluster use an existing scheduler is easy enough, just set the `scheduler` member to something that has the correct `address` attribute.

Allowing for multiple clusters to spin up and connect to the same cluster, for example one LocalCluster and one of remote workers, requires ensuring the workers get different names. The code in various dask repos assumes that worker names are integers, so I didn't want to change that (meaning when I did, exceptions were raised). Instead, this pull request adds two other arguments to set what the cluster will call its workers:
`workers_name_start` and `workers_name_step`.

The workers will be called integers starting from `workers_name_start`, with step size `workers_name_step`.

I have been using this code successfully in the following way:
1. Start a LocalCluster with a few workers, say 20 on an AWS c5n.metal.
2. If I want more workers, I call ECSCluster with `workers_name_start=1000`, and at some point the scheduler just starts passing tasks to the new workers.

This pull request addresses issue #322 .

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -498,6 +498,10 @@ class ECSCluster(SpecCluster):
         Defaults to `None`, no extra command line arguments.
     scheduler_task_kwargs: dict (optional)
         Additional keyword arguments for the scheduler ECS task.
+    scheduler_address: str (optional)
+        If passed, no scheduler task will be started, and instead the workers will connect to the passed address.
+
+        Defaults to `None`, a scheduler task will start.
     worker_cpu: int (optional)
         The amount of CPU to request for worker tasks in milli-cpu (1/1024).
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
Seems reasonable to me, thanks!

RichardScottOZ:
@drorspei workflow for computing results back to the local machine as opposed to ecs scheduler?

RichardScottOZ:
Could you pass to FargateCluster too?

PR #325, opened 2021-12-08
VMCluster should name wokers based on SpecCluster new_worker_name()

ashleysommer:
This was already working for new VMs created using the scale() feature of SpecCluster, but this change allows subclasses of VMCluster to specify their own _new_worker_name, and its initial n_workers (created in _start()) will be named accordingly.


ashleysommer:
Note, this could also be implemented using SpecCluster's `new_worker_spec`:

```
_ = [self.worker_spec.update(self.new_worker_spec() for i in range(self._n_workers)]
```
That would save some duplicated code between VMCluster and SpecCluster.

jacobtomlinson (approved):
Many thanks!

PR #326, opened 2021-12-15
Support for current version of az-* libraries

lejmr:
I came to the problem that get_azure_cli_credentials does not return any credentials after a succesfull `az login` command. This fixed the issue

### DIFF
--- a/dask_cloudprovider/azure/azurevm.py
+++ b/dask_cloudprovider/azure/azurevm.py
@@ -13,7 +13,7 @@ from dask_cloudprovider.generic.vmcluster import (
 from dask_cloudprovider.exceptions import ConfigError
 
 try:
-    from azure.common.credentials import get_azure_cli_credentials
+    from azure.common.credentials import get_cli_profile
     from azure.mgmt.network import NetworkManagementClient
     from azure.mgmt.compute import ComputeManagementClient
     from azure.identity import DefaultAzureCredential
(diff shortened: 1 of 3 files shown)


dask_cloudprovider/azure/azurevm.py
```diff
@@ -478,6 +479,7 @@ def __init__(
             "azurevm.public_ingress", override_with=public_ingress
         )
         self.credentials, self.subscription_id = get_azure_cli_credentials()
+        self.credentials = DefaultAzureCredential()
```
jacobtomlinson, line 482: Perhaps we could add a check for none here?

```suggestion
        if self.credentials is None:
            self.credentials = DefaultAzureCredential()
```

requirements.txt
```diff
@@ -2,4 +2,5 @@ aiohttp>=3.7.3
 dask>=2021.01.1
 distributed>=2021.01.1
 jinja2
-tornado>=5
\ No newline at end of file
+tornado>=5
+azure-identity
```
jacobtomlinson, line 6: These requirements are just the minimum for the package and not any of the providers. 

We use extra requirements to keep each cloud's dependencies separate.

So could you add it here instead?

https://github.com/dask/dask-cloudprovider/blob/f9c3272f833dad56101a9a8310039ac7cd5dc8a0/setup.py#L10-L15

jacobtomlinson (commented):
Thanks for raising this. A few comments.

dask_cloudprovider/azure/azurevm.py
lejmr: There was a slighly different problem related to the internal AZ logic change: https://docs.microsoft.com/en-us/cli/azure/msal-based-azure-cli

This should be more generic approach for receiving credentials and subscription_id

jacobtomlinson (approved):
Thanks for making these changes!

Issue #71, opened 2020-02-28
Problem with cleanup and region


rsignell-usgs:
For some reason cleanup is not recognizing the region:
```python
from dask_cloudprovider import FargateCluster
cluster = FargateCluster(n_workers=1, 
                         image='rsignell/pangeo-worker:2020-01-23c', region_name='us-west-2')
```
gives
```python-traceback
---------------------------------------------------------------------------
NoRegionError                             Traceback (most recent call last)
<timed exec> in <module>

~/SageMaker/myenvs/pangeo/lib/python3.7/site-packages/dask_cloudprovider/providers/aws/ecs.py in __init__(self, **kwargs)
   1099 
   1100     def __init__(self, **kwargs):
-> 1101         super().__init__(fargate_scheduler=True, fargate_workers=True, **kwargs)
   1102 
   1103 

~/SageMaker/myenvs/pangeo/lib/python3.7/site-packages/dask_cloudprovider/providers/aws/ecs.py in __init__(self, fargate_scheduler, fargate_workers, image, scheduler_cpu, scheduler_mem, scheduler_timeout, worker_cpu, worker_mem, worker_gpu, n_workers, cluster_arn, cluster_name_template, execution_role_arn, task_role_arn, task_role_policies, cloudwatch_logs_group, cloudwatch_logs_stream_prefix, cloudwatch_logs_default_retention, vpc, subnets, security_groups, environment, tags, find_address_timeout, skip_cleanup, aws_access_key_id, aws_secret_access_key, region_name, **kwargs)
    593         self._region_name = region_name
    594         self._lock = asyncio.Lock()
--> 595         super().__init__(**kwargs)
    596 
    597     async def _start(self,):

~/SageMaker/myenvs/pangeo/lib/python3.7/site-packages/distributed/deploy/spec.py in __init__(self, workers, scheduler, worker, asynchronous, loop, security, silence_logs, name)
    254         if not self.asynchronous:
    255             self._loop_runner.start()
--> 256             self.sync(self._start)
    257             self.sync(self._correct_state)
    258 

~/SageMaker/myenvs/pangeo/lib/python3.7/site-packages/distributed/deploy/cluster.py in sync(self, func, asynchronous, callback_timeout, *args, **kwargs)
    158             return future
    159         else:
--> 160             return sync(self.loop, func, *args, **kwargs)
    161 
    162     async def _logs(self, scheduler=True, workers=True):

[...]

PR #334, opened 2022-03-15
Pass appropriate kwargs to _cleanup_stale_resources for ECS

browningluke:
On creation of an ECS cluster, an error is thrown if `~/.aws/config` does not exist. It will also only check for stale resources in the default region specified in `~/.aws/config`.

I've updated `_cleanup_stale_resources` to allow passing of kwargs to fix this. `ECSCluster(region_name="xxxx")` will now check for stale resources in the specified region before creation.

This PR should also fix issue #71.


jacobtomlinson (approved):
Awesome stuff thanks!

CI failure is due to Dask dropping Python 3.7. I'll open another PR to resolve.

Issue #328, opened 2021-12-20
Fargate with spot instances fails with InvalidParameterException


tracek:
**What happened**:

Fargate (AWS) with spot instances fails with:

```
InvalidParameterException: An error occurred (InvalidParameterException) when calling the RunTask operation: The specified capacity provider strategy cannot contain a capacity provider that is not associated with the cluster. Associate the capacity provider with the cluster or specify a valid capacity provider and try again.
```

**What you expected to happen**:

Cluster creation succeeds and I get to enjoy spot instances on Fargate

**Minimal Complete Verifiable Example**:

```python
from dask_cloudprovider.aws import FargateCluster

cluster = FargateCluster(fargate_spot=True)
```

**Environment**:

- Dask version: dask cloudprovider: `2021.6.0+16.g876ccba` (master), dask `2021.12.0`
- Python version: 3.8, 3.9
- Operating System: Ubuntu 18.04
- Install method (conda, pip, source): source

<details>

Full stack trace:
```
---------------------------------------------------------------------------
InvalidParameterException                 Traceback (most recent call last)
/tmp/ipykernel_11179/3565260805.py in <module>
----> 1 cluster = FargateCluster(image="<redacted>", fargate_spot=True)

/home/ubuntu/miniconda3/envs/dask3/lib/python3.9/site-packages/dask_cloudprovider/aws/ecs.py in __init__(self, **kwargs)
   1460 
   1461     def __init__(self, **kwargs):
-> 1462         super().__init__(fargate_scheduler=True, fargate_workers=True, **kwargs)
   1463 
   1464 

[...]

PR #348, opened 2022-05-11
 Fargate with spot instances fails with InvalidParameterException #328

davidvhill:
ECS would not allow FARGATE_SPOT tasks to be created without FARGATE_SPOT being added as a capacityProvider to the cluster.

Also added FARGATE to support schedulers (on demand) and workers (spot) within the same cluster.


jacobtomlinson (approved):
This looks reasonable to me. Many thanks!

PR #355, opened 2022-06-03
Add support for GCP labels on instance creation

kapadia:
This PR is similar to #352 where AWS instances may be launched with user defined tags. In the GCP nomenclature these are called labels. These changes introduce support for user defined labels at launch time.

Please let me know if there are some changes that need to be made. If I recall correctly, there's some code formatting directives that might bark at me after this merge request is submitted.


jacobtomlinson (approved):
Awesome thanks

PR #357, opened 2022-06-27
ECSCluster attr cleanup

pwerth:
This PR adds `fargate_use_private_ip` to `cloudprovider.yaml`, which allows us to set defaults using dask config without passing a value to the `FargateCluster` constructor explicitly.

While making this update, I noticed that there is a lot of duplicated logic in `ECSCluster._start()`, specifically with how class attributes are set/updated. I cleaned this up by creating a helper function. This helper function is implemented as a mixin, as it would be nice to also use it in `EC2Cluster`. However, the corresponding logic there is a bit different. Namely, arguments from the constructor are not saved to the instance, and the config update happens in the constructor (as opposed to `_start`). So I decided not to include that in this PR, but I would be happy to refactor to use this mixin if you're ok with it @jacobtomlinson.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -697,9 +697,9 @@ class ECSCluster(SpecCluster, ConfigMixin):
 
     def __init__(
         self,
-        fargate_scheduler=False,
-        fargate_workers=False,
-        fargate_spot=False,
+        fargate_scheduler=None,
+        fargate_workers=None,
+        fargate_spot=None,
         image=None,
         scheduler_cpu=None,
         scheduler_mem=None,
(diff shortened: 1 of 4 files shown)


pwerth:
Hm, it turns out this is a bit more subtle than I realized. Because `fargate_use_private_ip` has a default value of `False` and not `None`, my `update_attr_from_config` function will not properly handle it. Interestingly, there are 3 such boolean attributes that are [referred to in the same way](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/aws/ecs.py#L829-L834). 

Is this a bug? If so, I can remove `fargate_scheduler`, `fargate_worker` and `fargate_spot` from the list of attributes that I loop over. Alternatively, I can set the default values of these arguments, along with that of `fargate_use_private_ip`, to be `None` instead of `False`. This has the benefit of allowing the defaults to be set directly on dask config before creating the cluster. The downside, I suppose, is that `None` is not a bool.

Would appreciate some guidance here. Converting this to draft until then.

jacobtomlinson:
Every kwarg should be loaded from the config. I expect these few situations where they are not are external contributions and I forgot to ask for the config to be used during review.

I'd be very happy for these defaults to be moved to the config.

pwerth:
> Every kwarg should be loaded from the config. I expect these few situations where they are not are external contributions and I forgot to ask for the config to be used during review.
> 
> I'd be very happy for these defaults to be moved to the config.

Great, I've just un-marked as draft. Apart from the refactor of attribute setting to use my new mixin, there are a few small logic and formatting changes:

1. Change default values of kwargs `fargate_scheduler`, `fargate_spot` and `fargate_workers` to be `None` instead of `False`. This fixes the bug I pointed out in my earlier comment - namely, that [these lines](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/aws/ecs.py#L829-L834) don't do anything because the variables will never be `None`. 
2. Add `fargate_spot` and `fargate_use_private_ip` to `cloudprovider.yaml`, per your comment.
3. Re-formatted `False` to `false` in `cloudprovider.yaml`. It seems we use `true` instead of `True` elsewhere in the file, so this was just a formatting change for consistency. 

I'd be happy to make a similar cleanup in `ec2.py`, but feel that would be better in a separate PR.

jacobtomlinson (approved):
Awesome thanks!

PR #358, opened 2022-06-30
Option to Provide Task Definition ARNs to ECSCluster

### LINKED ISSUE #135: Add ability to specify task definitions for scheduler and workers
From the code in ecs.py, I see that the task definitions for both schedulers and workers are created on the fly during cluster creation:

        self.scheduler_task_definition_arn = (
            await self._create_scheduler_task_definition_arn()
        )
        self.worker_task_definition_arn = (
            await self._create_worker_task_definition_arn()
        )

Would it be possible to specify both the scheduler/worker task definitions during the function call for creating the Fargate or ECS cluster?

WToma:
Aims to resolve #135.

This set of changes provides an option to provide task definitions to the `ECSCluster` rather than having the cluster create one on startup. This can help with the proliferation of task definition families on the AWS UI, and with providing various settings to the task definition that the cluster doesn't support directly.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -754,10 +754,6 @@ class ECSCluster(SpecCluster, ConfigMixin):
         task_role_policies=None,
         cloudwatch_logs_group=None,
         cloudwatch_logs_stream_prefix=None,
-        scheduler_logs_group=None,
-        scheduler_logs_prefix=None,
-        worker_logs_group=None,
-        worker_logs_prefix=None,
         cloudwatch_logs_default_retention=None,
         vpc=None,
         subnets=None,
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/aws/ecs.py
```diff
@@ -401,9 +401,17 @@ class Scheduler(Task):
     See :class:`Task` for parameter info.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, scheduler_timeout, scheduler_extra_args=None, **kwargs):
         super().__init__(**kwargs)
         self.task_type = "scheduler"
+        self._overrides = {
```
WToma, line 407: Added this so that the scheduler timeout can be properly set on the scheduler. Previously this argument was part of the task definition.

dask_cloudprovider/aws/ecs.py
```diff
@@ -904,35 +964,72 @@ async def _start(
...
 
-        if self._execution_role_arn is None:
-            self._execution_role_arn = (
-                self.config.get("execution_role_arn")
-                or await self._create_execution_role()
-            )
-
-        if self._task_role_policies is None:
-            self._task_role_policies = self.config.get("task_role_policies")
+        if (
```
WToma, line 967: This part is best viewed with whitespace changes hidden.

dask_cloudprovider/aws/ecs.py
```diff
@@ -904,35 +964,72 @@ async def _start(
...
             self._cloudwatch_logs_default_retention = self.config.get(
                 "cloudwatch_logs_default_retention"
             )
 
-        if self.cloudwatch_logs_group is None:
-            self.cloudwatch_logs_group = (
-                self.config.get("cloudwatch_logs_group")
-                or await self._create_cloudwatch_logs_group()
-            )
+        if self._scheduler_logs_prefix is None or self.scheduler_logs_group is None:
```
WToma, line 990: Suggestions are more then welcome about how to DRY this up in an idiomatic way.

Tl;dr: if the logging information is not configured, but the task definition was provided, retrieve the logging information from the task definition.

jacobtomlinson (commented):
Thanks for taking the time to raise this, I'm very happy to see this PR.

I'm generally happy with the direction this is going in. I am always hesitant around adding new kwargs and configuration options as cluster managers in this package are already very complex to configure correctly. However in this case I think it is very beneficial. 

Retrieving the address from the logs has always been a compromise. I would be very happy to see a PR that removes this.

In terms of PR hygiene things generally seem good, linting is passing and existing tests pass. We would need more tests and some documentation before this can be merged though.

tamas4sunairio:
(Using my work account, but I'm the same person who opened the PR :))

Sorry for the long delay on this PR! I added the extra documentation. Re: the tests, it seems that the current tests are just testing the imports. Am I missing something?

I also see that there's another PR (#364) which also seems stalled. I think this PR provides more options, but maybe the other one's advantage is that it's simpler. I'm OK with going forward with either!

tamas4sunairio:
Rebased from `main`, allowing to remove the extra complexity introduced for finding the log location.

PR #364, opened 2022-07-14
Parameterize task definition name for ECS/Fargate clusters

jkanche:


dask_cloudprovider/aws/ecs.py
```diff
@@ -254,7 +260,9 @@ async def start(self):
                         "overrides": {
                             "containerOverrides": [
                                 {
-                                    "name": "dask-{}".format(self.task_type),
+                                    "name": self.task_def_name
+                                    if self.task_def_name
+                                    else "dask-{}".format(self.task_type),
```
jacobtomlinson, line 265: Is this branch ever going to be run? It looks like the task name defaults to `dask-worker-{uuid}` further down?

dask_cloudprovider/aws/ecs.py
```diff
@@ -755,6 +768,11 @@ def __init__(
         self._worker_mem = worker_mem
         self._worker_gpu = worker_gpu
         self._worker_extra_args = worker_extra_args
+        self._worker_task_def_name = (
+            worker_task_def_name
+            if worker_task_def_name
+            else "dask-worker-{}".format(str(uuid.uuid4())[:5])
+        )
```
jacobtomlinson, line 775: I wonder if we can switch this up so we for the formatting after it chooses either the user config or the default. That was a user could pass a task def name like `"my-awesome-task-def-{uuid}"` and it will still be populated with a UUID.

```suggestion
        self._worker_task_def_name = (
            worker_task_def_name
            if worker_task_def_name
            else "dask-worker-{uuid}"
        ).format(uuid=str(uuid.uuid4())[:5])
        
```

jacobtomlinson (commented):
This is awesome thanks @jkanche! A couple of small comments.

dask_cloudprovider/aws/ecs.py
jkanche: that makes sense

dask_cloudprovider/aws/ecs.py
jkanche: in the current workflow probably not. I made a change so that the `task_def_name` is a bit more generalized. Can be reused for both worker and scheduler container definitions

dask_cloudprovider/aws/ecs.py
```diff
@@ -755,6 +768,16 @@ def __init__(
...
+        self._worker_task_def_name = (
+            str(task_def_name) + "-worker-{uuid}"
+            if task_def_name
+            else "dask-worker-{uuid}"
+        ).format(uuid=str(uuid.uuid4())[:5])
+        self._scheduler_task_def_name = (
+            str(task_def_name) + "-scheduler-{uuid}"
+            if task_def_name
+            else "dask-scheduler-{uuid}"
+        ).format(uuid=str(uuid.uuid4())[:5])
```
jacobtomlinson, line 780: Instead of setting these values here could you just set them to whatever the kwarg is and then handle them in `_start`. If you have a look at all the other options they are done this way.

Then in `_start` you'll need to check if they are `None` and if so load them from `dask.config`. All default config values are set in `cloudprovider.yaml`.

jacobtomlinson (commented):
Splitting out worker and scheduler names makes sense. Just one more comment about default values and using the Dask config system.

dask_cloudprovider/aws/ecs.py
```diff
@@ -953,6 +967,18 @@ async def _start(
                 or await self._create_security_groups()
             )
 
+        self._worker_task_def_name = (
+            f"{str(self._task_def_name)}-worker-{uuid}"
```
jacobtomlinson, line 971: I think you might've misunderstood my comment. This should just be the task def name, we should allow the user to construct whatever format string they like. If they place `{uuid}` in the string it will be replaced with the uuid.

```suggestion
            self._task_def_name
```

dask_cloudprovider/aws/ecs.py
```diff
@@ -953,6 +967,18 @@ async def _start(
...
+            if self._task_def_name
+            else "dask-worker-{uuid}"
+        ).format(uuid=str(uuid.uuid4())[:5])
```
jacobtomlinson, line 974: In other parts of Dask we also unpack `os.environ` into formats like these which allows users to incorporate environment variables into their config. For example they could set the task def name to `"dask-worker-task-{USER}-{uuid}"` to include their username.

```suggestion
        ).format(uuid=str(uuid.uuid4())[:5], **os.environ)
```

dask_cloudprovider/aws/ecs.py
jacobtomlinson: Sorry this hasn't been addressed yet. Marking as unresolved.

dask_cloudprovider/aws/ecs.py
```diff
@@ -953,6 +968,16 @@ async def _start(
                 or await self._create_security_groups()
             )
 
+        self._worker_task_def_name = (
+            self._task_def_name if self._task_def_name else "dask-worker-{uuid}"
+        ).format(uuid=str(uuid.uuid4())[:5], **os.environ)
+
+        self._scheduler_task_def_name = (
+            f"{str(self._task_def_name)}-scheduler-{uuid}"
```
jacobtomlinson, line 976: This still needs fixing up the same as the worker task name

jkanche: not sure I quite understand, if we don't add `scheduler` in there somewhere, the worker and scheduler task definitions would be the same and that would cause conflicts

jacobtomlinson:
I'm going to close this out in favour of #358.

PR #367, opened 2022-08-01
Destroy VM is creation fails to avoid leaking VMs

martinedgefocus:
1. Pulled out the postprocessing into configure_vm, and wrap in try/catch
2. Put create_tags in a timeout loop.
3. moved initialization of the backoff variable outside of the timeout loop for ip_address_key, else it seems pointless?
4. moved the sleep call to end of loop for ip_address_key - sleeping before attempt seems wrong?
5. changed the ip_address_key check to self.instance.get(ip_address_key) - simpler, more readable, and equivalent?
6. botocore.config.Config(retries = dict(max_attempts=10)) - without this we get errors, seems to be a fairly standard AWS problem.

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -162,10 +162,12 @@ class EC2Instance(VMInterface):
             response = await client.run_instances(**vm_kwargs)
             [self.instance] = response["Instances"]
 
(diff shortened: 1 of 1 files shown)


jacobtomlinson (approved):
This looks great, many thanks for raising this.

PR #368, opened 2022-08-02
Ability to enable detailed monitoring for EC2 instances

pwerth:
As title. See [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-cloudwatch-new.html) for documentation on detailed vs. basic monitoring.

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -55,7 +55,7 @@ class EC2Instance(VMInterface):
         instance_tags: None,
         volume_tags: None,
         use_private_ip: False,
-        enable_detailed_monitoring: False,
+        enable_detailed_monitoring=False,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -361,7 +361,7 @@ class EC2Cluster(VMCluster):
     enable_detailed_monitoring: bool (optional)
         Whether to enable detailed monitoring for created instances.
         See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-cloudwatch-new.html
-        Default ``False``.
+        Default ``None``.
 
     Notes
     -----
@@ -470,7 +470,7 @@ class EC2Cluster(VMCluster):
         instance_tags=None,
         volume_tags=None,
         use_private_ip=False,
-        enable_detailed_monitoring=False,
+        enable_detailed_monitoring=None,
         **kwargs,
     ):
         self.boto_session = get_session()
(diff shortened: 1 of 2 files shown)


jacobtomlinson (changes requested):
This looks great!

Could you change the default kwarg to `None` and load it from the `dask.config` for consistency with the other config options.

For example
https://github.com/dask/dask-cloudprovider/blob/3592253592b8aa8bf056542709c5cc2f2a3100de/dask_cloudprovider/aws/ec2.py#L473

pwerth:
> This looks great!
> 
> Could you change the default kwarg to `None` and load it from the `dask.config` for consistency with the other config options.
> 
> For example
> 
> https://github.com/dask/dask-cloudprovider/blob/3592253592b8aa8bf056542709c5cc2f2a3100de/dask_cloudprovider/aws/ec2.py#L473

Done! I also added the default to `cloudprovider.yaml`. 

FYI I have a [draft PR](https://github.com/dask/dask-cloudprovider/pull/357) to streamline a lot of the kwarg/config parsing in `ECSCluster`; something similar could probably be done in `EC2Cluster` as well. Not sure if you've seen that PR but do you think this is a worthwhile change? I can close the PR if you think it's not necessary.

jacobtomlinson:
> Done! I also added the default to cloudprovider.yaml.

Perfect thanks!

> Not sure if you've seen that PR but do you think this is a worthwhile change?

Apologies it had slipped past my radar. It looks great and I've commented over there.

dask_cloudprovider/aws/ec2.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         instance_tags: None,
         volume_tags: None,
         use_private_ip: False,
+        enable_detailed_monitoring=False,
```
jacobtomlinson, line 58: Should this also be `None` like the rest?

dask_cloudprovider/aws/ec2.py
```diff
@@ -356,6 +358,10 @@ class EC2Cluster(VMCluster):
         Whether to use a private IP (if True) or public IP (if False).
 
         Default ``False``.
+    enable_detailed_monitoring: bool (optional)
+        Whether to enable detailed monitoring for created instances.
+        See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-cloudwatch-new.html
+        Default ``None``.
```
jacobtomlinson, line 364: The docstring here should reflect the actual default, so from the config this would be ``False``.

jacobtomlinson (approved):
This is great thanks!

PR #369, opened 2022-08-09
EC2Cluster support for different instance types on scheduler and workers

pwerth:
`FargateCluster` supports different instance types for the scheduler and workers (via the `scheduler_mem`, `scheduler_cpu`, `worker_mem`, `worker_cpu` arguments), but `EC2Cluster` does not provide this flexibility. This PR changes this by introducing two new arguments to `EC2Cluster`: `scheduler_instance_type` and `worker_instance_type`. 

I believe this flexibility gives a better experience, and it also establishes consistency with `FargateCluster`. For these reasons I propose deprecating `instance_type` in favor of the new args above. This PR was written with this deprecation in mind, however I did not want to introduce a breaking change. Thus my solution is to raise a deprecation warning when users supply the `instance_type` argument, and in this case both the scheduler and workers share that type (as they currently do). However, an error is raised if users supply `instance_type` _and_ one of the new arguments. Default values are not changed, so users who don't pass in any of these args will see no change.

I'm open to feedback on my assumptions above. Hopefully I didn't come in too strong with a deprecation without opening an issue first 🙃

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -516,7 +516,7 @@ class EC2Cluster(VMCluster):
             if instance_type is not None
             else self.config.get("instance_type")
         )
-        if self.instance_type is None:
+        if instance_type is None:
             self.scheduler_instance_type = (
                 scheduler_instance_type
                 if scheduler_instance_type is not None
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -533,7 +533,8 @@ class EC2Cluster(VMCluster):
                     "If you specify instance_type, you may not specify scheduler_instance_type or worker_instance_type"
                 )
             warnings.warn(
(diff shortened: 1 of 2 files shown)


jacobtomlinson (commented):
I think adding these additional options is a great idea, folks do occasionally want to put the scheduler on a smaller instance.

I definitely consider this to be an advanced configuration option though and I think many users will want to use the same type for both. So I'm not sure if we should deprecate the old option. Raising if the old and new are both set makes sense, but I don't think we should stop users from continuing to use `instance_type` to set both if they don't want to make that distinction.

pwerth:
> I think adding these additional options is a great idea, folks do occasionally want to put the scheduler on a smaller instance.
> 
> 
> 
> I definitely consider this to be an advanced configuration option though and I think many users will want to use the same type for both. So I'm not sure if we should deprecate the old option. Raising if the old and new are both set makes sense, but I don't think we should stop users from continuing to use `instance_type` to set both if they don't want to make that distinction.

Got it - so keep all 3 then? And I'll remove the deprecation warning.

jacobtomlinson (approved):

PR #371, opened 2022-08-10
use gp3 by default for AWS EBS storage

martinedgefocus:
gp3 is cheaper and more performant than gp2
https://aws.amazon.com/blogs/storage/migrate-your-amazon-ebs-volumes-from-gp2-to-gp3-and-save-up-to-20-on-costs/


jacobtomlinson (approved):
Awesome!

PR #372, opened 2022-08-17
Enable all EC2 config options in `cloudprovider.yaml`

pwerth:
This PR makes 2 updates with the goal of enabling more configuration to be done "behind the scenes" rather than supplied explicitly to the `EC2Cluster` constructor.

1. Fixes a bug (introduced in #353) where `EC2Cluster` constructor argument `use_private_ip` had incorrect default value, preventing intended behavior w.r.t loading from `cloudprovider.yaml` default. This argument was not actually in `cloudprovider.yaml` before, so this PR adds it as well.
2.  Uncomment  `ami`, `vpc`, `subnet_id` and `security_groups` from `cloudprovider.yaml` EC2 config. To test my change, I ran the following snippet and verified that the created cluster had the correct ami; was in the VPC, subnet, and security_groups specified in my config, and used private IP address.

```
import dask
from dask_cloudprovider.aws import EC2Cluster

my_dask_config = {
    "cloudprovider": {
        "ec2": {
            "security_groups": ["my-security-group-id"],
            "subnet_id": "my-subnet-id",
            "use_private_ip": True,
            "vpc": "my-vpc-id",
        },
    },
}

dask.config.update(dask.config.config, my_dask_config)

ec2_cluster = EC2Cluster(
    security=False, # https://github.com/dask/dask-cloudprovider/issues/249,
    n_workers=2
)
```

### DIFF
--- a/dask_cloudprovider/aws/ec2.py
+++ b/dask_cloudprovider/aws/ec2.py
@@ -484,7 +484,7 @@ class EC2Cluster(VMCluster):
         debug=False,
         instance_tags=None,
         volume_tags=None,
-        use_private_ip=False,
+        use_private_ip=None,
         enable_detailed_monitoring=None,
         **kwargs,
     ):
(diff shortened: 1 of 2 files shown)


jacobtomlinson (approved):
Thanks for another high-quality PR! Really appreciate your efforts here.

Issue #381, opened 2022-09-30
Remove IAM actions from minimal policy for AWS FargateCluster


jgdwyer:
In the interest of minimizing the number of permissions that we need to provide to the cluster manager, I'd like to explore the option of removing the need for `iam:ListRoles` and `iam:ListRoleTags` [actions](https://github.com/dask/dask-cloudprovider/blob/9f5e73047a082cb322c50c2ec0d6cd33b8950d31/dask_cloudprovider/aws/ecs.py#L1471-L1472) if all resources are specified when creating the cluster. Looking at the code, it appears that if the `task_role_arn` and `execution_role_arn` are specified, then the iam actions are only needed when attempting to [clean up stale resources](https://github.com/dask/dask-cloudprovider/blob/9f5e73047a082cb322c50c2ec0d6cd33b8950d31/dask_cloudprovider/aws/ecs.py#L1565-L1584).

Is there support for skipping the cleanup of stale IAM roles if `task_role_arn` and `execution_role_arn` are explicitly specified? It might enable us to reduce the resources needed to ec2, ecs, and logs. (And for FargateCluster we could eliminate the need for ec2 as well).
```
"ec2:CreateTags",
"ec2:DescribeNetworkInterfaces",
"ec2:DescribeSecurityGroups",
"ec2:DescribeSubnets",
"ec2:DescribeVpcs",
"ecs:DescribeTasks",
"ecs:ListAccountSettings",
"ecs:RegisterTaskDefinition",
"ecs:RunTask",
"ecs:StopTask",
"ecs:ListClusters",
"ecs:DescribeClusters",
"ecs:ListTaskDefinitions",
"ecs:DescribeTaskDefinition",
"ecs:DeregisterTaskDefinition",
"logs:DescribeLogGroups",
"logs:GetLogEvents"
```

jgdwyer:
> Cleaning up stale resources can be skipped today, are you suggesting automatically skipping if everything is provided?

I was thinking to automatically skip the role cleanup if the roles were provided, but to keep the rest of the automatic cleanup in place. But perhaps there are other cleanup steps that can be skipped as well if their resources are directly provided?

jgdwyer:
Yup I can put something together!

PR #385, opened 2022-10-20
Minimize aws permissions required for ecs/fargate cluster

jgdwyer:
Intended to address https://github.com/dask/dask-cloudprovider/issues/381

### Overview
The goal is to reduce the number of AWS actions/permissions required to run the ECSCluster or FargateCluster. While one can pass `skip_cleanup=True` to avoid needing many permissions, sometimes one may want cleanup to happen on relevant resources. For example, if one always passes specific iam task roles when launching the cluster, there's no need for the cluster to try to clean up stale iam task roles. By only cleaning up resources that the cluster creates on its own, we can reduce the minimal permissions set.

### Changes
- Add conditional statements to only attempt to clean resources if resource arn's have not been provided for them
- Move the `_cleanup_stale_resources` to be a class method rather than a standalone function
- Break up `_cleanup_stale_resources` into helper functions for readability purposes

### Testing
I've not attempted to test this at all yet, so leaving this as a draft PR for now. But if you all are on board with this kind of approach, I will do some manual testing on my end.

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -1343,7 +1343,6 @@ class ECSCluster(SpecCluster, ConfigMixin):
                 "port."
             )
 
-
     async def _cleanup_stale_resources(self):
         """Clean up any stale resources which are tagged with 'createdBy': 'dask-cloudprovider'.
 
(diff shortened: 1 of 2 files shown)


jacobtomlinson:
This has been in draft for a very long time, so I'm going to close it out. Thanks for the effort here, sorry it never got merged.

PR #387, opened 2022-10-25
ECS: Add Option to Specify TLS Scheduler

### LINKED ISSUE #386: Change to Scheduler Address Parsing Breaks TLS Encryption Support
We're still investigating more but we believe this recent PR https://github.com/dask/dask-cloudprovider/pull/373 broke Dask Cloud Provider provisioned clusters on AWS that use TLS encryption. Previously, when the scheduler address was parsed from logs, it would correctly parse the url prefix of `tls://` to indicate the Dask client should use TLS. With the PR, no prefix seems to be included for the scheduler address (the new code in the PR appears to be just `return f"{ip}:{self.port}"`) so the Dask client defaults to assuming `tcp://` for the scheduler address and then fails with:

```
RuntimeError: encryption required by Dask configuration, refusing communication from/to 'tcp://172.29.39.222:8786'
```

Note that the Dask scheduler started with this log output:

```
2022-10-24 12:59:10,760 - distributed.scheduler - INFO - Scheduler at: tls://172.29.39.222:8786
2022-10-24 12:59:10,760 - distributed.scheduler - INFO - dashboard at: :8787
```

We pass the following kwargs to FargateCluster to configure TLS encryption with constants for the Security object and filepaths:

```
        security=SRM_DOCKER_DASK_TLS_SECURITY,
        scheduler_extra_args=[
            "--tls-cert",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_client_cert,
            "--tls-key",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_client_key,
            "--tls-ca-file",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_ca_file,
        ],
        worker_extra_args=[
            "--tls-cert",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_client_cert,
            "--tls-key",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_client_key,
            "--tls-ca-file",
            SRM_DOCKER_DASK_TLS_SECURITY.tls_ca_file,
        ],
```

If you need additional details on environment, etc. please let me know, but this should be easy to recreate and we don't think it's related to anything specifically in our environment. Thanks!

tamas4sunairio:
Adds an option to specify that the scheduler is on TLS
Fixes #386

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -502,13 +502,6 @@ class ECSCluster(SpecCluster, ConfigMixin):
         The port on which the scheduler should listen.
 
         Defaults to ``8786``
-    scheduler_tls: bool (optional)
-        Whether or not the scheduler is expecting TLS connections.
-
-        Defaults to ``false``.
-        Note: the value of this argument needs to be consistent with the TLS arguments provided in
-        `scheduler_extra_args`, otherwise a warning will be logged, and the workers won't be able
-        to connect to the cluster.
     scheduler_extra_args: List[str] (optional)
         Any extra command line arguments to pass to dask-scheduler, e.g. ``["--tls-cert", "/path/to/cert.pem"]``
 
@@ -722,7 +715,6 @@ class ECSCluster(SpecCluster, ConfigMixin):
         scheduler_cpu=None,
         scheduler_mem=None,
         scheduler_port=8786,
-        scheduler_tls=False,
         scheduler_timeout=None,
         scheduler_extra_args=None,
         scheduler_task_definition_arn=None,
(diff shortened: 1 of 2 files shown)


jacobtomlinson (changes requested):
Thanks for taking the time to raise time.

The class already takes a `security=` keyword argument (you can find it on the base class).
https://github.com/dask/distributed/blob/621994ec4484eb690dba796951f8b365f4bcd247/distributed/deploy/spec.py#L228

I would prefer to reuse this option because it is also set by the `dask.config` system too. So instead of taking `scheduler_tls` as a boolean, you should e able to just check if `self.security` is `None`.

tamas4sunairio:
Makes sense, updated. With this, does the `_check_scheduler_tls_config` function make sense, or should that be reworked/removed?

jacobtomlinson (approved):
Thanks for responding to the review feedback. It seems like a useful addition, many thanks!

Issue #173, opened 2020-11-17
Support multiple workers per VM


eric-czech:
It may be more efficient to run many workers on larger VMs as opposed to larger numbers of worker processes on small VMs, presumably to avoid inter-VM communication.  This was a suggestion that resulted from this thread on optimizing some dask array workflows: https://github.com/pystatgen/sgkit/issues/390.

@quasiben mentioned that threads per worker can be controlled with `worker_options={"nthreads": 2 }`, but there appears to be no way to run more than one worker on a single VM.

quasiben:
I think for this to work we need to change:

https://github.com/dask/dask-cloudprovider/blob/d5dfd99e0d29b1c6b03cb291633484fdd367ed4d/dask_cloudprovider/gcp/instances.py#L318-L330

To handle a spec which looks like the following:


```python
workers_options = {
    'worker-1': {"cls": "dask.distributed.Nanny", "opts": {"nthreads": 1}},
    'worker-2': {"cls": "dask.distributed.Nanny", "opts": {"nthreads": 2}},
}
```

jacobtomlinson:
When running on a local machine `LocalCluster` optimises this for you right? On my laptop with 12 cores it runs three workers with four threads each.

Perhaps we can reuse this logic?

quasiben:
You mean, have dask-cloudprovider inspect the system and assign worker/threads accordingly ?

jacobtomlinson:
Yeah. We do this for `LocalCluster` [here](https://github.com/dask/distributed/blob/eda9bcca8baa502481bf929f39340857da1ebf09/distributed/deploy/local.py#L246-L265).

jacobtomlinson:
dask/distributed#4377 allows us to detect and create worker/threads automatically.

```
dask-worker --nprocs=auto
```

Once a release is out we can update things here to use this new option.

eric-valente:
any progress here?

jacobtomlinson:
@eric-valente with the latest release of `dask` and `distributed` you should be able to set `worker_options={"nprocs": "auto"}`. Although I haven't managed to come back here and test this yet.

If you could give it a go and report back it would be much appreciated!

eric-valente:
Hi @jacobtomlinson  Tried it with packages and no luck.   The workers will turn on in EC2 and won't connect to the scheduler and then proceed to die.  When I remove the nprocs it works fine but again only has 1 worker per VM with threads = # of cores 

[...]

PR #389, opened 2022-11-08
add new param to allow multiple workers in vmcluster, worker_mixin

kumarprabhu1988:
Intended to address [#173](https://github.com/dask/dask-cloudprovider/issues/173)

**Overview**
The goal is to have the ability to spin up mulitple workers on VM Clusters (like `EC2Cluster`). 

- A new param `num_workers_per_vm` is added to the `VMCluster` and `WorkerMixin` classes. 
- This param defaults to 1 in both classes (this is the current behavior). 
- `num_workers_per_vm` worker classes are added to the spec (which are passed to `SpecCluster`) to initialize multiple workers.
- To keep changes minimal, there is no validation of this parameter to ensure `num_workers_per_vm` is less than number of cores available. 

**Testing**
Confirmed this works by creating an `EC2Cluster` as follows

```
ec2_cluster = EC2Cluster(
    env_vars=env_vars,
    extra_bootstrap=EC2_BOOTSTRAP_COMMANDS,
    filesystem_size=cluster_config["volume_size"],
    instance_tags=cluster_config["ec2_instance_tags"],
    n_workers=10,
    worker_class="distributed.nanny.Nanny",
    worker_options={"nthreads": 2, "memory_limit": "7GiB"},
    num_workers_per_vm=4,
    scheduler_instance_type=cluster_config["scheduler_instance_type"],
    auto_shutdown=False,
    shutdown_on_close=False,
    security=False,  # https://github.com/dask/dask-cloudprovider/issues/249,
    volume_tags=cluster_config["ec2_instance_tags"],
    worker_instance_type="m5.2xlarge",
)
```
creates an `EC2Cluster` with 10 m5.2xlarge worker machines and a total of 40 workers (as `num_workers_per_vm` = 4).

### DIFF
--- a/dask_cloudprovider/generic/vmcluster.py
+++ b/dask_cloudprovider/generic/vmcluster.py
@@ -101,7 +101,7 @@ class WorkerMixin(object):
         worker_module: str = None,
         worker_class: str = None,
         worker_options: dict = {},
-	num_workers_per_vm: int = 1,
+        num_workers_per_vm: int = 1,
         **kwargs,
     ):
         super().__init__(*args, **kwargs)
@@ -124,7 +124,7 @@ class WorkerMixin(object):
             )
         if worker_class is not None:
             self.worker_class = worker_class
-	    self.num_workers_per_vm = num_workers_per_vm
+            self.num_workers_per_vm = num_workers_per_vm
             self.command = " ".join(
                 [
                     self.set_env,
@@ -135,15 +135,15 @@ class WorkerMixin(object):
                     "--spec",
                     "''%s''"  # in yaml double single quotes escape the single quote
                     % json.dumps(
-			{
-			    f"{i}": {
-				"cls": self.worker_class,
-				"opts": {
-				    **worker_options,
-				    "name": self.name,
-				},
-			    }
-			    for i in range(self.num_workers_per_vm)
+                        {
+                            f"{i}": {
+                                "cls": self.worker_class,
+                                "opts": {
+                                    **worker_options,
+                                    "name": self.name,
+                                },
+                            }
+                            for i in range(self.num_workers_per_vm)
                         }
                     ),
                 ]
@@ -229,7 +229,7 @@ class VMCluster(SpecCluster):
         n_workers: int = 0,
         worker_class: str = "dask.distributed.Nanny",
         worker_options: dict = {},
-	num_workers_per_vm: int = 1,
+        num_workers_per_vm: int = 1,
         scheduler_options: dict = {},
         docker_image="daskdev/dask:latest",
         docker_args: str = "",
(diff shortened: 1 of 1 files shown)


jacobtomlinson (commented):
This seems great, I'm happy it is such a minimal change.

Would you be able to go through the docstrings of the classes that implement the `VMCluster` interface and document this option?

Also could you add a test?

kumarprabhu1988:
@jacobtomlinson Thanks for the quick review. Sure, I can add testa and update docstrings. Do have a question for tests. Quickly looked at existing tests and it seems like the right place to put this test would be in `EC2Cluster` tests, rather than `VMCluster` tests. Is this correct?

jacobtomlinson:
Yeah that sounds good.

erl987:
Did you have any progress on this pull request in recent time?

This option would be very helpful as well for the `GCPCluster`. Most machine types on GCP have more than one CPU - this makes `dask-cloudprovider` wasting resources currently.

erl987:
I tested this new option `num_workers_per_vm` and it does not seem to work for `GCPCluster`. This code reproduces the problem:

```python
import time
from dask_cloudprovider.gcp import GCPCluster
from dask.distributed import Client

cluster = GCPCluster(n_workers=1, num_workers_per_vm=2, worker_options={'nthreads': 1}, projectid='the-project-id')
time.sleep(180)
client = Client(cluster)
print(client)
cluster.close()
```

The printout is `<Client: 'tls://10.142.0.11:8786' processes=1 threads=1, memory=3.59 GiB>`, i.e., only one worker got created.

jacobtomlinson:
> Most machine types on GCP have more than one CPU - this makes dask-cloudprovider wasting resources currently.

I'm not sure this is true. the default behaviour should start 1 thread per CPU core.

erl987:
Maybe, but I have an application that is not thread-safe due to the underlying FORTRAN code being used.

jacobtomlinson:
Ah fair enough. I'm not sure I understand why this following isn't suitable though.

```python
cluster = GCPCluster(n_workers=1, worker_options={'nthreads': 1, 'nworkers': 2}, projectid='the-project-id')
```

erl987:
According to the documentation, this option does not exist for workers, only threads: https://distributed.dask.org/en/latest/worker.html#distributed.worker.Worker

Trying it also fails.

jacobtomlinson:
Ah yeah I remember, in `dask-cloudprovider` we use the `dask spec` CLI to start the worker instead of `dask worker` which is why we can't do that. 

To get this PR over the line we need:
- Tests
- Docs
- Update to use the Dask config system
- Maybe rename `num_workers_per_vm` to something more consistent with other options like `n_worker_procs`

Given that this seems abandoned @erl987 do you have any interest in pushing it forwards?

flbulgarelli:
:wave: Hi @jacobtomlinson! I indeed have interest in continuing this PR, since it is something we are needing at https://github.com/CaviDBOrg/. 

After struggling for a few days last week with the same problem, I found this PR and  I am currently testing those changes on AWS. In parallel, I have created a new branch based on this one - https://github.com/flbulgarelli/dask-cloudprovider/commits/feature-allow-multiple-workers-in-vm-cluster - that address some of your requests, but I am still a bit unsure about your expectations about testing it. 

As far as I understand, all of the [tests here](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/aws/tests/test_ec2.py) are smoke tests, while others do actually run distributed computations at some point. I am thinking of starting a cluster and checking the actual number of workers `using client.scheduler_info()['workers']`. Do you have any other recommendations?

 Thanks!

jacobtomlinson:
There has been no activity here for a long time, so I'm going to close it out. Thanks for the effort here, sorry it never got merged.

Issue #254, opened 2021-02-09
AzureVMCluster workers unable to connect to scheduler


arh89:
**What happened**:
I'm just getting started with dask-cloudprovider, and looking to spin up an `AzureVMCluster` to parallelise my dask workloads. I followed the [documentation](https://cloudprovider.dask.org/en/latest/azure.html), to create the required infrastructure (resource group, network security group, vnet) except that, rather than allowing network traffic from the internet to ports 8786-8787, I restricted the IP to be the one of my local machine.

Following the minimal example in the documentation, i.e:
```python
```

**What you expected to happen**:

**Minimal Complete Verifiable Example**:

```python
# Put your MCVE code here
```

**Anything else we need to know?**:

**Environment**:

- Dask version:
- Python version:
- Operating System:
- Install method (conda, pip, source):

quasiben:
It's possible the workers are trying to communicate with the scheduler using a public IP address rather than an internal IP addr.  I am not as familiar with Azure, but this is the case for GCP and AWS.  There is a `public_ingress` option which probably forces the dask cluster to be in the VPC.  

Taking a step back, your desired configuration is a publicly addressable scheduler and dashboard with workers in the VPC, correct ?

arh89:
Hi @quasiben, thanks for taking the time to respond.

> It's possible the workers are trying to communicate with the scheduler using a public IP address rather than an internal IP addr.

Yes, I think you might be right with that one.

> There is a public_ingress option which probably forces the dask cluster to be in the VPC.

Looking at the documentation: `public_ingress: bool` `Assign a public IP address to the scheduler. Default True.` Looks like this just creates the public IP – I guess I could set this to `False` and assign the public ip to the scheduler manually through the Azure portal. I'll give that a try and see how I get on...

> Taking a step back, your desired configuration is a publicly addressable scheduler and dashboard with workers in the VPC, correct ?

Yep – that's right. 👍

quasiben:
We have had similar discussions on the GCP: https://github.com/dask/dask-cloudprovider/issues/229 https://github.com/dask/dask-cloudprovider/issues/215 .  

And we track internal ips from GCP:
https://github.com/dask/dask-cloudprovider/blob/dffd81d3e4ab6feb5378ffa0d9952f3c7da207f7/dask_cloudprovider/gcp/instances.py#L302-L305

[...]

PR #394, opened 2022-12-09
Ensure workers use private ip to connect to scheduler

nordange:
Intended to address [#254](https://github.com/dask/dask-cloudprovider/issues/254)

**Overview**
Let the scheduler and dashboard be publicly adressable with the workers in a vnet. The workers will always connect to the scheduler using the private IP, regardless of whether the scheduler has a public IP or not.

- `self.address` is the private IP
- `self.external_address` is the public IP on the scheduler object if public_ingress=True

**Testing**
Confirmed that it works by creating a cluster with `public_ingress = True` and a vnet allowing traffic on ports 8786-8787 from IP of local laptop only. Without this fix, the workers try to communicate with the scheduler using the public IP.

### DIFF
--- a/dask_cloudprovider/azure/azurevm.py
+++ b/dask_cloudprovider/azure/azurevm.py
@@ -14,8 +14,6 @@ from dask_cloudprovider.generic.vmcluster import (
 from dask_cloudprovider.exceptions import ConfigError
 from dask_cloudprovider.azure.utils import _get_default_subscription
 
-from distributed.core import Status
-
 try:
     from azure.mgmt.network import NetworkManagementClient
     from azure.mgmt.compute import ComputeManagementClient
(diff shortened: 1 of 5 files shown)


dask_cloudprovider/generic/vmcluster.py
```diff
@@ -43,7 +43,10 @@ async def destroy_vm(self):
         raise NotImplementedError("destroy_vm is a required method of the VMInterface")
 
     async def wait_for_scheduler(self):
-        _, address = self.address.split("://")
+        if getattr(self, "external_address", None):
```
jacobtomlinson, line 46: I would expect `external_address` to always exist so it would probably be more readable to do:

```suggestion
        if self.external_address:
```

dask_cloudprovider/azure/azurevm.py
```diff
@@ -241,10 +246,44 @@ async def destroy_vm(self):
 class AzureVMScheduler(SchedulerMixin, AzureVM):
     """Scheduler running on an Azure VM."""
 
+    def __init__(self, *args, **kwargs):
+        kwargs.pop("preemptible", None)
+        super().__init__(*args, **kwargs)
+
+    async def start(self):
```
jacobtomlinson, line 253: Instead of overriding the `start` methods I would prefer to update the mixins themselves and expect `create_vm` to return a tuple. For clouds where this isn't being implemented yet they can just `return address, None`.

jacobtomlinson (changes requested):
Thanks! This seems like a great enhancement and we should be doing this everywhere. Instead of special casing Azure I'd rather implement this in a way that allows us to implement it for other clouds easily in the future. See my commments.

jacobtomlinson (approved):
Awesome thanks. Sorry for the delay in coming back here.

jacobtomlinson:
Pulled from `main` to hopefully fix up linting failures.

PR #399, opened 2023-02-02
Replace legacy commands with 'dask worker' and 'dask scheduler'.

wilson:
Hi there.
I'm not 100% the change I made re: GPU workers is correct.
I believe it to be, but I haven't used it in anger.
Looking at the `dask-cuda` project, it does look like `dask-cuda-worker` is deprecated just like `dask-worker`.

Anyway, this PR fixes a pair of warnings:
```
distributed/cli/dask_scheduler.py:140: FutureWarning: dask-scheduler is deprecated and will be removed in a future release; use `dask scheduler` 
distributed/cli/dask_worker.py:264: FutureWarning: dask-worker is deprecated and will be removed in a future release; use `dask worker` instead
```

In doing this, I realized that the `worker_command` option is not used anywhere, so this PR also removes the remaining references to it.


jacobtomlinson (commented):
This look great thanks.

In `dask-cuda` we replaced `dask-cuda-worker` with `dask cuda worker`. That's all that should be changed. Sorry you went down the resources path, that's not needed.

wilson:
OK, rebased off `main` and all looks green.

dask_cloudprovider/aws/ecs.py
```diff
@@ -434,24 +435,25 @@ def __init__(
         self._mem = mem
         self._gpu = gpu
         self._nthreads = nthreads
+        _command = [
+            "dask",
+            "cuda" if self._gpu else None,
```
jacobtomlinson, line 440: I'm apprehensive about having a `None` in this string. Have you tested that this works as expected?

dask_cloudprovider/aws/ecs.py
```diff
@@ -434,24 +435,25 @@ def __init__(
         self._mem = mem
         self._gpu = gpu
         self._nthreads = nthreads
+        _command = [
```
jacobtomlinson, line 438: I'm curious why create this variable?

dask_cloudprovider/aws/ecs.py
```diff
@@ -1259,17 +1262,23 @@ async def _create_worker_task_definition_arn(self):
...
-                                else self._worker_nthreads
-                            ),
-                            "--memory-limit",
-                            "{}MB".format(int(self._worker_mem)),
-                            "--death-timeout",
-                            "60",
+                            e
+                            for e in [
+                                "dask",
+                                "cuda" if self._worker_gpu else None,
```
jacobtomlinson, line 1268: Same here

jacobtomlinson (commented):
Thanks. Couple of questions.

dask_cloudprovider/aws/ecs.py
wilson: To both the above, just so it was more visible that we're filtering Nones out.
That seemed like the easiest way to have a variable number of arguments in the command.

dask_cloudprovider/aws/ecs.py
wilson: There's a None but it gets filtered out before we use it.

jacobtomlinson: I found the introduction and filtering of `None` hard to parse when reviewing this.

I think it would be more readable to have something like

```python
if self._gpu:
    _command = ["dask", "cuda", "worker"]
else:
    _command = ["dask", "worker"]
_command += [OTHERARGS...]
```

jacobtomlinson:
There's been no activity here for a while so I'm going to close this PR out. @wilson if you have the desire to pick this up I'd happily give it another review.

PR #413, opened 2023-07-20
Ability to specify service account to run VMs in GCP.

### LINKED ISSUE #403: [GCP] Set custom service account for worker VMs
# Requested functionality
**Example use case:** The workers need to download data from Google Cloud Storage. They require the *Cloud Storage Viewer* role for this purpose.

I would like to set a custom service account for the worker VMs that contains this role.

# Current state
As far as I can see, no service account can be set for the worker or scheduler. Instead, I need to add a service account key file (having the required role) to the Docker container of the worker / scheduler and make the environment variable `GOOGLE_APPLICATION_CREDENTIALS` pointing to it in the container. Then for example the Python package `google-cloud-storage` can access the storage buckets from the worker.

This is undesirable, both regarding flexibility and security.

jmhatze:
Adds a new parameter, `service account`, to the constructor of `GCPCluster` which allows the user to specify a service account to run VMs under in GCP. This parameter can also be specified in the dask config file. Defaults to the default Compute Engine service account for the GCP project in question.

Closes #403.

### DIFF
--- a/dask_cloudprovider/cloudprovider.yaml
+++ b/dask_cloudprovider/cloudprovider.yaml
@@ -108,6 +108,7 @@ cloudprovider:
     public_ingress: true # configure the scheduler to be externally accessible.  This assumes firefwall rules for 8787 and 8786
     instance_labels:
       container_vm: "dask-cloudprovider"
+    service_account: "default"
 
   hetzner:
     token: null # API token for interacting with the Hetzner cloud API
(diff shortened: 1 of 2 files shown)


jmhatze:
Not sure why those three tests are failing/cancelled, but it does not look related to my changes as the same issues arise in the commit 7b214adc87f1026d8e2149fa109276f6e2497606 in main.

jacobtomlinson (approved):
Apologies for the delay in reviewing this. This looks great thanks.

PR #416, opened 2023-10-18
Fix CI failing tests

lukaszo:
It looks like CI currently fails because of three reasons:

1. **pytest-asyncio** defaults now to the [strict mode](https://pytest-asyncio.readthedocs.io/en/latest/concepts.html#strict-mode) 
There are two ways to fix it. Switch back to using auto mode or replace all `@pytest.fixture` with `@pytest_asyncio.fixture`
In this PR I've used the first option. If you prefer the second option, I will update the PR.
2. **dask** dropped support for python3.8
When creating conda env for 3.8 it fails with the following error message:
`ERROR: Package 'dask' requires a different Python: 3.8.18 not in '>=3.9'` 
To fix it, I've removed python3.8 and updated all examples to use 3.9
3. Outdated Read The Docs config.
Updated.


lukaszo:
Even though I've removed it, it looks like it's still waiting for the 3.8 test to finish. Maybe it's some kind of protection as I'm not a member of the team, but I'm modifying the CI  :man_shrugging:

lukaszo:
Ok, it seems that the test for 3.8 is marked in `Require status checks to pass before merging` list in the repo settings.

lukaszo:
Hi @jacobtomlinson Will you have a time to check this out?

jacobtomlinson (approved):
Sorry for the delay in reviewing here. This looks great. At some point I would like to explore dropping `pytest-asyncio` in favour of `pytest-anyio` with just the `asycnio` backend as we have done in other parts of Dask, but in the meantime this gets things going.

I'll change the branch protection rules accordingly.

PR #417, opened 2023-10-27
Add Fly.io as a provider

mileszim:
Hello! This PR adds the cloud platform host [Fly.io](https://fly.io) as a provider. It includes the auto-generated documentation too.

Let me know if anything is missing. Thank you!


dask_cloudprovider/__init__.py
```diff
@@ -42,3 +42,9 @@ def __getattr__(name):
             "DigitalOcean cluster managers must be imported from the digitalocean subpackage. "
             f"Please import dask_cloudprovider.digitalocean.{name}"
         )
+
+    if name in ["FlyMachineCluster"]:
+        raise ImportError(
+            "Fly.io cluster managers must be imported from the fly subpackage. "
+            f"Please import dask_cloudprovider.fly.{name}"
+        )
```
jacobtomlinson, line 50: These shims only exist for historic reasons because previously some of these classes were available at the top level. I don't see a need to add a shim for a new provider.

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -117,3 +117,13 @@ cloudprovider:
     image: "ubuntu-20.04" # Operating System image to use
     docker_image: "daskdev/dask:latest" # docker image to use
     bootstrap: true # It is assumed that the OS image does not have Docker and needs bootstrapping. Set this to false if using a custom image with Docker already installed.
+
+  fly:
+    token: null # API token for interacting with the Fly API
+    region: "lax" # Region to launch Droplets in
+    vm_size: "shared-cpu-1x" # Droplet size to launch, default is 1GB RAM, 1 vCPU
+    image: "daskdev/dask:latest" # Operating System image to use
```
jacobtomlinson, line 125: This should use GHCR. We should probably update the others too.
```suggestion
    image: "ghcr.io/dask/dask:latest" # Operating System image to use
```

setup.py
```diff
@@ -15,6 +15,7 @@
     "digitalocean": ["python-digitalocean>=1.15.0"],
     "gcp": ["google-api-python-client>=1.12.5", "google-auth>=1.23.0"],
     "hetzner": ["hcloud>=1.10.0"],
+    "fly": ["httpx>=0.24.0", "pydantic>=1.10.7"],
```
jacobtomlinson, line 18: Shouldn't this include `fly-python-sdk`?

jacobtomlinson (commented):
Thanks for contributing this, I've never heard of fly.io. I'm happy to merge this in but I don't think I'll be much help in maintaining it so if issues come up I'll ping you on them.

I noticed you copied in the whole SDK rather than adding it as a dependency. Can I ask why?

mileszim:
> Thanks for contributing this, I've never heard of fly.io. I'm happy to merge this in but I don't think I'll be much help in maintaining it so if issues come up I'll ping you on them.
> 
> I noticed you copied in the whole SDK rather than adding it as a dependency. Can I ask why?

I'm not in control over that repo, and it isn't clear how much if any effort will be spent in the future by the maintainer to update or fix issues, so I figured the most reliable thing to do, especially for such a small subset of the Fly API, is to implement those calls myself

jacobtomlinson:
Just want to check in here and see if there is any interest in continuing with this PR?

PR #418, opened 2023-11-01
Bump flake8 and black in pre-commit config and add support for 3.11

lukaszo:
This fixes pre-commit failure https://github.com/dask/dask-cloudprovider/actions/runs/6709474379/job/18232600682

It looks like the default python version in the action was bumped to the latest 3.12. **flake8** in version `3.9.2` does not work with python3.12.
I've bumped the **flake8** and **black** to the latest versions. The new **black** formatter removed a few empty lines.

I've also added a support for python 3.11 and preliminary support for 3.12. Before supporting 3.12 we have to wait for a new aiohttp version https://github.com/aio-libs/aiohttp/issues/7675

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: true
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
 
     steps:
       - name: Checkout source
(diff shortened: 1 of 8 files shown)


lukaszo:
I also had to set python version to 3.11 when building docs. It seems that everyone is upgrading to python 3.12 :)

lukaszo:
@jacobtomlinson ping :)

jacobtomlinson (approved):
Thanks for your patience here. This looks great.

PR #424, opened 2024-01-18
Add example of authenticating Docker in VMs to use GCP Artifact Registry

nprihodko:
Without additional steps GCPCluster cannot use images from Artifact Registries in the same project. Add example of how to set up access.


jacobtomlinson (approved):
Awesome thanks!

PR #425, opened 2024-03-12
ARM64 support for ECS and Fargate

dmitry-livchak:
Support explicit runtime platform architecture selection for ECS and Fargate clusters

### DIFF
--- a/dask_cloudprovider/aws/ecs.py
+++ b/dask_cloudprovider/aws/ecs.py
@@ -759,11 +759,11 @@ class ECSCluster(SpecCluster, ConfigMixin):
         mount_volumes_on_scheduler=False,
         **kwargs,
     ):
-        self._cpu_architecture = cpu_architecture
         self._fargate_scheduler = fargate_scheduler
         self._fargate_workers = fargate_workers
         self._fargate_spot = fargate_spot
         self.image = image
+        self._cpu_architecture = cpu_architecture.upper()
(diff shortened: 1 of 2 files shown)


dask_cloudprovider/aws/ecs.py
```diff
@@ -754,6 +759,7 @@ def __init__(
         mount_volumes_on_scheduler=False,
         **kwargs,
     ):
+        self._cpu_architecture = cpu_architecture
```
jacobtomlinson, line 762: Could we add a case check here?

It looks like valid values are `X86_64` and `ARM64`. However the lowercase `x86_64` and `arm64` are also very commonly used in other situations (this is what `uname -m` will show).

Perhaps we want to either explicitly set this with `.upper()` or check the value against known good values (although if more are added in the future this adds a maintenance burden).

jacobtomlinson (commented):
Thanks for raising this!

dask_cloudprovider/aws/ecs.py
dmitry-livchak: .upper() it is!

dmitry-livchak: @jacobtomlinson I also moved it up in line with order of the constructor fields. Happy to proceed?

jacobtomlinson (approved):
This is great, thank you! The RTD failure is unrelated so happy to merge.

PR #434, opened 2024-09-05
Add support for IBM Cloud

macarronesc:


dask_cloudprovider/__init__.py
```diff
@@ -42,3 +42,8 @@ def __getattr__(name):
             "DigitalOcean cluster managers must be imported from the digitalocean subpackage. "
             f"Please import dask_cloudprovider.digitalocean.{name}"
         )
+    if name in ["IBMCodeEngineCluster"]:
+        raise ImportError(
+            "IBM cluster managers must be imported from the ibm subpackage. "
+            f"Please import dask_cloudprovider.ibm.{name}"
+        )
```
jacobtomlinson, line 49: I wouldn't worry about this. These shims were added when we rearranged the package at one point. But given that nobody h as ever imported `IBMCodeEngineCluster` from this location I don't think we need it.

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -119,6 +119,18 @@ cloudprovider:
     docker_image: "daskdev/dask:latest" # docker image to use
     bootstrap: true # It is assumed that the OS image does not have Docker and needs bootstrapping. Set this to false if using a custom image with Docker already installed.
 
+  ibm:
+    api_key: null
+    image: "daskdev/dask:latest"
```
jacobtomlinson, line 124: We probably need to do this everywhere, but this image is no longer maintained and we should be using the GitHub hosted ones.

```suggestion
    image: "ghcr.io/dask/dask:latest"
```

dask_cloudprovider/cloudprovider.yaml
```diff
@@ -119,6 +119,18 @@ cloudprovider:
     docker_image: "daskdev/dask:latest" # docker image to use
     bootstrap: true # It is assumed that the OS image does not have Docker and needs bootstrapping. Set this to false if using a custom image with Docker already installed.
 
+  ibm:
+    api_key: null
```
jacobtomlinson, line 123: Is there not a global location where IBM stores credentials? For example AWS stores things in `~/.aws/credentials` and we read it from there.

jacobtomlinson (commented):
Thanks for taking the time here!

I don't have an IBM Cloud setup to test this against, but I'm happy to merge it in provided you've verified things work as expected.

As this contribution is a [puppy](https://textual.textualize.io/blog/2023/07/29/pull-requests-are-cake-or-puppies/) I'd appreciate your commitment to continue maintaining this code and respond to issues and fix bugs in this integration. Are you happy to do that?

dask_cloudprovider/cloudprovider.yaml
macarronesc: No, according to the documentation it is not stored anywhere visible to the user. That is why it is necessary to add the API key manually.

macarronesc:
I have tested the Code Engine backend on IBM and everything works fine.

Of course! I'll be happy to help fix bugs and suggestions that may arise over time in this backend.

jacobtomlinson (approved):
Awesome many thanks!

Issue #438, opened 2024-10-04
Plugins that help to pass credentials for S3 and GCS to remote cluster workers


dbalabka:
I didn't find a simple way to pass credentials to remote workers, such as S3 and GCS, while both are widely used to store data frames.
In this ticket's scope, I propose creating plugins that will help distribute the required keys to remote workers.

**GCP credentials**
GCP credentials file path is stored in `GOOGLE_APPLICATION_CREDENTIALS` env variable. The plugin has to create a remote file and pass an env variable with a proper path to workers.

**S3 credentials**
Like GCP, we must update credential files and store them on each worker.

jacobtomlinson:
Usually you would create an [IAM instance role and profile](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) that can access S3, then configure workers to have this role via the `iam_instance_profile` keyword argument. 

The GCP equivalent is to create a service account that can access GCS and configure that with the `service_account` kwarg. 

This way you don't have to pass credentials around. Is there a reason why you aren't doing it this way?

dbalabka:
@jacobtomlinson, sorry for not being active the last few months because of workload and vacation. Thanks for the question.

You are correct that using a proper service account or IAM role/profile is the more secure way and preferable for production workloads. However, I have a few scenarios when dynamically uploading the key can be more convenient.

For **local development**, the recommended way for GCP cloud is to use Application Default Credentials acquired with `gcloud auth application-default login` command. Previously, I provided changes and a detailed description in #429. ADC is associated with developers [user account](https://cloud.google.com/docs/authentication#user-accounts) that would be preferable to reuse in workers. Otherwise, we have to create a separate service account key for each developer or automate the creation of such a key. 

If dask deployed **on-prem Kubernetes** during local development, uploading the key is the most convenient way to provide the key to workers. Otherwise, we have to keep them in [Secrets ](https://kubernetes.io/docs/concepts/configuration/secret/) and mount them separately. However, such an approach is more suitable for production workloads.

jacobtomlinson:
I see. So we could create a plugin for the client which grabs those credentials and propagates them to the workers. Do you have any interest in implementing such a plugin?

dbalabka:
[...]

PR #439, opened 2024-10-04
Create credentials upload plugins for GCP and AWS (#438)

dbalabka:


dask_cloudprovider/aws/plugins.py
```diff
@@ -0,0 +1,57 @@
+import logging
+import os
+from pathlib import Path
+
+from distributed import WorkerPlugin
+
+logger = logging.getLogger(__name__)
```
dbalabka, line 7: I have to add some logging for the visibility of the fact that file has been uploaded

dask_cloudprovider/gcp/plugins.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+from pathlib import Path
+
+from distributed import WorkerPlugin
+from google.auth._cloud_sdk import get_application_default_credentials_path
+
+logger = logging.getLogger(__name__)
+
+class UploadGCPKey(WorkerPlugin):
+    """Automatically upload a GCP key to the worker."""
```
dbalabka, line 11: Need to add more context about why we need to upload the key from the laptop.

dask_cloudprovider/aws/plugins.py
```diff
@@ -0,0 +1,57 @@
...
+class UploadAWSCredentials(WorkerPlugin):
+    # """Automatically upload a GCP key to the worker."""
+
+    name = "upload_aws_credentials"
+
+    def __init__(self):
+        """
+        Initialize the plugin by reading in the data from the given file.
+        """
+        config_path = os.getenv("AWS_CONFIG_FILE", Path.home() / Path(".aws/config"))
```
dbalabka, line 18: Should why upload the complete config?

dbalabka (commented):
I need to improve the code some. I will get back to it 😀

jacobtomlinson (commented):
Overall this seems like a great addition. This definitely needs some testing and documentation (see the [Azure spot termination plugin](https://cloudprovider.dask.org/en/latest/azure.html#azure-spot-instance-plugin) for an example of how we documented that).

I'm marking this as a draft as it looks like you still intend to do more work here. Give me a ping when this is ready for review.

dbalabka:
@jacobtomlinson, thanks for the good reference code. I will let you know when it can be reviewed.

PR #442, opened 2024-12-31
[aws] add docs on cpu_architecture for ECSCluster

jameslamb:
Using `FargateCluster` tonight, I found myself wanting to create a cluster using arm64 servers, because I was working with container images that were only built from arm64.

Thankfully, as of #425, that's now configurable easily via the `cpu_architecture` argument 🎉 

... but I didn't know what the valid values were for that. `arm64`? `aarch64`? does capitalization matter?

This proposes adding a link to the relevant AWS docs to help answer that question.

## Notes for Reviewers

Tested `cpu_architecture="ARM64"` tonight and it worked great 😁 

ref: https://github.com/jameslamb/lightgbm-dask-testing/pull/68


jacobtomlinson (approved):
Many thanks!

PR #443, opened 2025-01-21
Add Nebius AI Cloud support

SalikovAlex:
Add [Nebius AI Cloud](https://nebius.com) support

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,6 +17,7 @@ extras_require = {
     "hetzner": ["hcloud>=1.10.0"],
     "ibm": ["ibm_code_engine_sdk>=3.1.0"],
     "openstack": ["openstacksdk>=3.3.0"],
+    "nebius": ["nebius @ git+https://github.com/nebius/pysdk"]
 }
 extras_require["all"] = set(pkg for pkgs in extras_require.values() for pkg in pkgs)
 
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -17,6 +17,7 @@
     "hetzner": ["hcloud>=1.10.0"],
     "ibm": ["ibm_code_engine_sdk>=3.1.0"],
     "openstack": ["openstacksdk>=3.3.0"],
+    "nebius": ["nebius @ git+https://github.com/nebius/pysdk"],
```
jacobtomlinson, line 20: Is there any road to getting this properly published on PyPI. Just pointing to the tip of main on GitHub is an uncomfortable place to be.

jacobtomlinson (commented):
Thanks for taking the time to work on this. I haven't heard of Nebius before, but it looks interesting.

This PR needs documentation, but other than that it looks good.

We don't have the capacity to maintain many different backends so I would ask that if we merge this that you can handle triaging bugs and making future PRs @SalikovAlex. If this is merged and then not maintained we will eventually to remove it again.

setup.py
SalikovAlex: I will fix it as soon as Nebius publishes lib on PiPy.

jacobtomlinson (commented):
Ok great thanks. Happy to merge once docs are added.

jacobtomlinson (approved):
Looks great thanks!

PR #446, opened 2025-05-13
[Openstack cloud provider] Fixed the floating ip logic

armagankaratosun:
Hey all 👋

I noticed that the floating IP assignment logic I initially implemented was faulty (since it was using an older API endpoint). I fixed the implementation, and it works as expected with floating IPs.

Cheers!


armagankaratosun:
@jacobtomlinson to review

jacobtomlinson (approved):
Looks good thanks!

PR #456, opened 2025-06-07
Remove calls to `self.cluster.call_async()`

jacobtomlinson:
Currently the implementations of `VMInterface` call `self.cluster.call_async()` when they want to run a synchronous function. This reference to the parent cluster instance means that `VMInterface` instances can't exist outside of a cluster.

This PR adds a `VMInterface.call_async` method which encapsulates the call to `self.cluster.call_async()`. This allows others to override this method if they wish more easily.


jacobtomlinson:
Ah ok, can you open a new issue to track this?

Issue #390, opened 2022-11-10
Support different machine types for scheduler and workers on GCP


jrbourbeau:
In https://dask.discourse.group/t/different-machine-types-for-scheduler-workers/1300 someone was asking for the ability to specify different machine types for their scheduler and workers on GCP. IIUC this isn't current implemented for `GCPCluster` but is for other managers (e.g. `EC2Cluster`). This seems like a reasonable feature request, so I wanted to log it here so we don't loose it.

jacobtomlinson:
Yeah this was merged for the AWS `EC2Cluster` in #369. If someone wants to contribute the same for GCP it should be a good reference.

PR #465, opened 2025-06-24
Fixing the log error

gmiasnychenko:
While working on the issue 390 in [this PR](https://github.com/dask/dask-cloudprovider/pull/451), it appears I have made a mistake with the logging "No GPU instances configured" message. The GCPCluster does not have any loggers, so it is better to just make a print statement.

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -650,7 +650,7 @@ class GCPCluster(VMCluster):
             self.scheduler_ngpus = scheduler_ngpus if scheduler_ngpus is not None else self.config.get("scheduler_ngpus", 0)
             self.worker_ngpus = worker_ngpus if worker_ngpus is not None else self.config.get("worker_ngpus", 0)
             if self.scheduler_ngpus == 0 and self.worker_ngpus == 0:
-                self._log("No GPU instances configured")
+                print("No GPU instances configured")
         else:
             if scheduler_ngpus is not None or worker_ngpus is not None:
                 raise ValueError("If you specify ngpus, you may not specify scheduler_ngpus or worker_ngpus")
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -649,8 +649,6 @@ class GCPCluster(VMCluster):
         if not self.ngpus:
             self.scheduler_ngpus = scheduler_ngpus if scheduler_ngpus is not None else self.config.get("scheduler_ngpus", 0)
             self.worker_ngpus = worker_ngpus if worker_ngpus is not None else self.config.get("worker_ngpus", 0)
-            if self.scheduler_ngpus == 0 and self.worker_ngpus == 0:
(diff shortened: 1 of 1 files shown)


gmiasnychenko:
The issue I encountered is related to the initialization order in the class hierarchy.

The problem occurs because the `self._log()` call) runs in `GCPCluster.__init__()` before `super().__init__()` is called. The `_log()` method depends on attributes (`_cluster_manager_logs` and `quiet`) that are initialized in the parent Cluster class.

The error I was getting:
```
...
  File "/home/user/miniconda3/envs/demand-forecasting/lib/python3.11/site-packages/ecentria_dask/gcp.py", line 547, in get_cloud_init
    return GCPCluster.get_cloud_init(**cluster_params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/demand-forecasting/lib/python3.11/site-packages/dask_cloudprovider/generic/vmcluster.py", line 378, in get_cloud_init
    cluster = cls(*args, asynchronous=True, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/demand-forecasting/lib/python3.11/site-packages/dask_cloudprovider/gcp/instances.py", line 653, in __init__
    self._log("No GPU instances configured")
  File "/home/user/miniconda3/envs/demand-forecasting/lib/python3.11/site-packages/distributed/deploy/cluster.py", line 302, in _log
    self._cluster_manager_logs.append((datetime.datetime.now(), log))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'GCPCluster' object has no attribute '_cluster_manager_logs'
```

I see the following way to solve this:
- Use the `print` statement instead of `self._log()`. As you said, it is not the best practice, but it keeps the logic in one place
- Defer the logging until after parent initialization. That would look something like:
```
...
if self.scheduler_ngpus == 0 and self.worker_ngpus == 0:
    gpu_log_message = "No GPU instances configured"

# ... rest of setup ...
super().__init__(debug=debug, **kwargs)
    
# Now we can safely use self._log()
if gpu_log_message:
    self._log(gpu_log_message)
```
In general, I think this is okay, but personally, it seems nicer to keep the logic in one place.
- Or, we can use standard logging at this point instead of `self._log()`:
```
import logging
logger = logging.getLogger(__name__)

def __init__(self, ...):
    # GPU configuration with standard logger
    if self.scheduler_ngpus == 0 and self.worker_ngpus == 0:
        logger.info("No GPU instances configured")  # Use standard logger during init
    
    # ... rest of setup ...
    super().__init__(debug=debug, **kwargs)
```

Please let me know which approach is better to proceed with, and I will adjust the PR accordingly.

gmiasnychenko:
Thanks for the feedback! I updated the PR in this manner. 

Also, maybe you can now close [the issue](https://github.com/dask/dask-cloudprovider/issues/390)?

Issue #249, opened 2021-01-28
Dask config size limitation in EC2Cluster


jacobtomlinson:
It seems there is a 16kb limit on the amount of user_data that can be passed to an EC2 instance when starting up.

We serialize the local Dask config and pass it to the scheduler and workers via the user_data. 

https://github.com/dask/dask-cloudprovider/blob/da454827b88c2f0f0d06af07e02d5d1580a4c366/dask_cloudprovider/generic/vmcluster.py#L33-L35

Depending on what config the user has locally this can tip us over the limit and result in the AWS API rejecting the instance creation call.

```python-traceback
botocore.exceptions.ClientError: An error occurred (InvalidParameterValue) when calling the RunInstances operation: User data is limited to 16384 bytes
```

douglasdavis:
I've run into this, after digging around a bit with @martindurant we've noticed that `dask.config.config` is modified (and grows quite a bit with only default settings) upon importing `dask_cloudprovider` or `dask_cloudprovider.{aws,gcp}`. (reproducer below).

Would it be reasonable to drop the keys in the config dict that have defaults? (especially things that are in the config only to spin up cloud resources, since the cloud instances don't need to know how to start things on the cloud?)

```python
In [1]: import dask.config

In [2]: dask.config.config
Out[2]: 
{'temporary-directory': None,
 'dataframe': {'shuffle-compression': None},
 'array': {'svg': {'size': 120}, 'slicing': {'split-large-chunks': None}},
 'optimization': {'fuse': {'active': True,
   'ave-width': 1,
   'max-width': None,
   'max-height': inf,
   'max-depth-new-edges': None,
   'subgraphs': None,
   'rename-keys': True}}}

In [3]: import dask_cloudprovider.aws

In [4]: dask.config.config
Out[4]: 
{'temporary-directory': None,
 'dataframe': {'shuffle-compression': None},
 'array': {'svg': {'size': 120}, 'slicing': {'split-large-chunks': None}},
 'optimization': {'fuse': {'active': True,
   'ave-width': 1,
   'max-width': None,
   'max-height': inf,
   'max-depth-new-edges': None,
   'subgraphs': None,
   'rename-keys': True}},
 'cloudprovider': {'ecs': {'fargate_scheduler': False,
   'fargate_workers': False,
   'scheduler_cpu': 1024,
   'scheduler_mem': 4096,
   'worker_cpu': 4096,
   'worker_mem': 16384,
   'worker_gpu': 0,
[...]

PR #467, opened 2025-07-02
Add pruning default dask configs before serialization

dshkliarenko:
Hi, this solves the issue (https://github.com/dask/dask-cloudprovider/issues/249). Thanks.

### DIFF
--- a/dask_cloudprovider/generic/vmcluster.py
+++ b/dask_cloudprovider/generic/vmcluster.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 import os
 import uuid
+import copy
 
 from jinja2 import Environment, FileSystemLoader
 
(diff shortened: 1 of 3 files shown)


jacobtomlinson (commented):
Thanks for picking up a 4 year old issue, this is awesome.

Could I ask you to write some tests for it?

dshkliarenko:
@jacobtomlinson I had to make some refactoring so that it could be covered with tests properly. 

Also, there is a caveat regarding the solution: it only works when security set to False, otherwise user data still is over limit because of serialized certificates.

jacobtomlinson:
Yeah this is a known issue. I don't think this PR actually loses #249 because most people want `security=True`. But it certainly help!

jacobtomlinson (approved):

PR #468, opened 2025-07-03
Fix build

dbalabka:
- [x] - pre-commit fix
- [x] - docs requirements instalation issue

Solutions:
1. `pre-commit` action was outdated and relied on old version of cache
2. docs requerements constrants are too broad, they should be more restrictive to avoid problem during dependecy resolution


jacobtomlinson (approved):
🎉 thank you!

PR #476, opened 2025-10-01
Allow for the specification of auth scopes for google compute engine instances

jakenotjay:
- Adds the ability to specify authentication oauth2 scopes for an instance in google compute engine
- This allows for workers/schedulers to use their credentials to access other services e.g. artifact registry or secret manager for any work


jacobtomlinson (changes requested):
Broadly this looks great.

Instead of setting the defaults in code could you move them into the [config](https://github.com/dask/dask-cloudprovider/blob/main/dask_cloudprovider/cloudprovider.yaml).

PR #477, opened 2025-10-03
Enable/disable public ingress at class instantiation for GCP cluster

jakenotjay:
- Adds the public ingress parameter to the GCP cluster class
- Allows for direct configuration instead of a hidden config item

### DIFF
--- a/dask_cloudprovider/gcp/instances.py
+++ b/dask_cloudprovider/gcp/instances.py
@@ -67,6 +67,7 @@ class GCPInstance(VMInterface):
         instance_labels=None,
         service_account=None,
         instance_scopes=None,
+        public_ingress=None,
         **kwargs,
     ):
         super().__init__(**kwargs)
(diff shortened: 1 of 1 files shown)


dask_cloudprovider/gcp/instances.py
```diff
@@ -107,6 +108,7 @@ def __init__(
         self.general_zone = "-".join(self.zone.split("-")[:2])  # us-east1-c -> us-east1
         self.service_account = service_account or self.config.get("service_account")
         self.instance_scopes = instance_scopes or self.config.get("instance_scopes")
+        self.public_ingress = public_ingress or self.config.get("public_ingress", True)
```
jacobtomlinson, line 111: Again can we put this default value in the config system? It just makes it easier for folks using the YAML to find the available configuration options.

jakenotjay: Hey jacob, I made this edit as it was already in the config, but wasn't accessible to users creating the cluster programatically see https://github.com/dask/dask-cloudprovider/blob/ca25d153d8c2ee85fb6a931b63663333dc468bc7/dask_cloudprovider/cloudprovider.yaml#L73

